Issue #2, opened 2018-01-03
Add middleware that configures the composition provider name in JSON


warpech:
Add middleware that configures the composition provider namespace according to the API introduced in https://github.com/Starcounter/starcounter-include/pull/80

warpech:
Done in PR #3.

Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Blending/releases/tag/5.0.0
- published in the App Warehouse

PR #3, opened 2018-01-04
Configure starcounter-include (for #2)

warpech:
In contrast to the original expectation, it is not possible at the moment to configure a static value of the namespace name for compositions provided by BlendingProvider, because it differs depending on the position in the merged JSON tree.

The index is 0 if the JSON subtree represents anything else than a Surface.
The index is 1 if the JSON subtree represents a Surface.

I have implemented it in form of a getter. Should work fine.

Implements #2.


src/BlendingProvider/Composition/Api/Middleware.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        public static void Register()
+        {
+            Application.Current.Use((request, response) =>
+            {
+                if (request.Uri == "/sys/starcounter-include/starcounter-include.html")
+                {
+                    var configuration = @"<script>
+                                    (function () {
+                                        var constructor = customElements.get('starcounter-include');
```
tomalec: To be picky, maybe `const`, and some comments would do the job.

src/BlendingProvider/Composition/Api/Middleware.cs
```diff
@@ -0,0 +1,37 @@
...
+                                        if (typeof constructor !== ""undefined"") {
+                                            Object.defineProperty(constructor.prototype, 'compositionProvider', { 
+                                                get: function() {
+                                                    return this.partial.BlendingProvider_1 && 'BlendingProvider_1' || 'BlendingProvider_0';
```
tomalec: To be bulletproof we could check `this.partial &&` first, however currently it's not needed and sometimes redundant.

warpech: Maybe it is keep it unchecked?

warpech:
Fixed according to https://github.com/Starcounter/Blending/pull/3#discussion_r159620542.

PatrykSzwer:
I've tested this with SCF 2.1.0 downloaded from the warehouse.
Manually everything seems to work perfectly, but I hoped that with this change tests for compositionProvider will start passing, but unfortunately they don't. Seems like starcounter-include doesn't recognize BlendingProvider correctly during tests session. I know that for tests entire DB is restarted, but I've tried to manually run SCF first, then all needed test projects, but still, without results.

warpech:
Since it works manually, I will merge it and then investigate the tests.

warpech:
- released (with release notes) https://github.com/Starcounter/Blending/releases/tag/5.0.0

Issue #25, opened 2018-01-08
⏰ Many tests don't pass


un-tone:
I didn't find an issue about that so I'm creating the new one.

Also, TeamCity doesn't run tests.

### Steps to reproduce

1. Build Blending solution
2. Run `test.bat`

### Expected result

All tests pass.

### Actual result

Many errors.

### Screenshot

Composition tests:
![image](https://user-images.githubusercontent.com/1625378/34674855-8b56672c-f498-11e7-83fd-6d6934f1982a.png)

Blending tests:

![image](https://user-images.githubusercontent.com/1625378/34674877-9a97cb40-f498-11e7-8ba1-33e90c8cf9d0.png)

warpech:
Last time I checked, all of the tests were passing in Chrome. Maybe you can try that? `--test.bat params="Browsers=Chrome"`

warpech:
The tests that fail in Firefox due to `InternalError: Too much recursion` seem to be caused by https://github.com/mozilla/geckodriver/issues/904.

Solution: change the helper function in Selenium to not return an element (`return arguments[0].shadowRoot`). The helper function should return only primitive values like string, number, boolean.

@PatrykSzwer can you take this?

warpech:
The tests that fail in Firefox due to `InternalError: Too much recursion` seem to be caused by https://github.com/mozilla/geckodriver/issues/904.

Solution: change the helper function in Selenium to not return an element (`return arguments[0].shadowRoot`). The helper function should return only primitive values like string, number, boolean.

KitchenSink is also affected by this.

@PatrykSzwer can you take this?

cc @alshakero

PatrykSzwer:
I've made PR fixing `Too much recursion` in Kitchensink: https://github.com/Starcounter/KitchenSink/pull/223. After acceptation, I'll apply referenced solution for Blending app.

warpech:
Closing as done in https://github.com/Starcounter/Blending/pull/33

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Blending/releases/tag/5.2.0
- published in the App Warehouse

PR #33, opened 2018-01-10
Fix for firefox failing tests

PatrykSzwer:
For #25 

With this fix, I've run all tests manually from VS level and every one of them passed.


test/BlendingProvider.Tests/Ui/BasePage.cs
```diff
@@ -67,15 +56,31 @@ public bool CheckForSidebarSurface(IWebElement starcounterInclude = null)
...
-        
+
         public void FillInput(IWebElement inputElement, string input)
         {
             inputElement.Clear();
             inputElement.SendKeys(input);
             inputElement.SendKeys(Keys.Enter);
         }
+
+        private IReadOnlyCollection<object> GetShadowObjects(IWebElement shadowElement, string markerClassName)
```
warpech: Wouldn't it be more maintainable if all our apps had the same API for BasePage.cs? https://github.com/Starcounter/KitchenSink/blob/eb41ea1971a7afbb4304d51c877e124fba60d3ec/test/KitchenSink.Tests/Ui/BasePage.cs#L109-L129

test/CompositionProvider.Tests/UI/BasePage.cs
```diff
@@ -27,16 +28,21 @@ public void ClickOn(IWebElement elementName, int seconds = 10)
             element.Click();
         }
 
-        public IWebElement ExpandShadowRoot(IWebElement shadowRootElement)
+        public IReadOnlyCollection<IWebElement> GetShadowElements(string markerClassName)
```
warpech: Wouldn't it be more maintainable if all our apps had the same or similar API for BasePage.cs? https://github.com/Starcounter/KitchenSink/blob/eb41ea1971a7afbb4304d51c877e124fba60d3ec/test/KitchenSink.Tests/Ui/BasePage.cs#L109-L129

PatrykSzwer: I'm not sure about maintainability of it, I've done It like that, because there is no other use case than looking at shadowRoot of element found by this: `Driver.FindElements(By.CssSelector("starcounter-include"))[1];`. If You want it for eventual future usage I could change it, but wouldn't it be more sufficient to left it like this since there is no need for more, and change it to more general method when It will be necessary?

test/BlendingProvider.Tests/Ui/BasePage.cs
PatrykSzwer: I need this method in this form because there were situations when returned by script entities couldn't be cast at IWebElement.

Issue #6, opened 2018-01-04
ScErrTransactionClosed after clicking tabs in specific order


PatrykSzwer:
This issue was moved from https://github.com/Starcounter/Website/issues/298

### Steps to reproduce
1. Go to `http://localhost:8080/WebsiteEditor`
2. Click two times at tab `Surface groups`
3. Click on `All routes` tab.
### Expected result
No errors
### Actual result
![image](https://user-images.githubusercontent.com/25042256/34200554-d1490f72-e571-11e7-90f3-0d26e88f5d59.png)

##### Starcounter version 2.3.1.8061

cc: @un-tone

un-tone:
Currently, the problem https://github.com/Starcounter/Blending/issues/25 impeds working on this issue.

I found out that currently the issue appears with enabled View Keeper and can be fixed in this line https://github.com/Starcounter/Blending/blob/d0f6147783f1febefb5cb21eb7ae6026a39ee3ed/src/BlendingProvider/Website/ViewModels/ViewKeeper.json.cs#L67 in the following way:

```cs
keptView.OriginalResponse = json;
```

But I'm not sure if it doesn't break something because of #25. 

Last time in Website repo with SC 2.3.1 I fixed it in the following way which also currently works here:

```cs
keptView.OriginalResponse = new Json();
keptView.OriginalResponse.MergeJson(json);
```

-----

What is currently required:

1. Get working tests.
2. Find a way to reproduce the issue in a test or a separate app.

warpech:
I am sorry for #25, we will work on improving that.

Before that happens, maybe you could continue your investigation in the Website repo, where there is a smaller amount of tests and all of them pass?

un-tone:
@warpech sure, I didn't stop working :-)

un-tone:
It's interesting that if I upgrade [`juicy-select`](https://github.com/Juicy/juicy-select) component to the last version the error is gone in this place. Removing `juicy-select` element from https://github.com/Starcounter/Blending/blob/c4af6afdd0eaa93765b5684ae2f35f6296f851f9/src/BlendingEditor/wwwroot/BlendingEditor/views/RoutingPage.html#L130-L133 make the same effect.

But the error appears again when I press  "Create" button:

![image](https://user-images.githubusercontent.com/1625378/34728089-f8a0bca2-f569-11e7-9bb7-5fd348623439.png)

warpech:
[...]

Issue #31, opened 2018-01-10
⏰InvalidCastException: Unable to cast object exception


PatrykSzwer:
### Steps to reproduce
1. Run Blending app.
2. Call `/BlendingEditor/resetdata`
3. Go to `/BlendingEditor/routing`
4. Select route for TopbarSurface and `Disable` it.
5. Go back to surfaceGroups and select TopbarSurface.
6. Go to routes tab.
7. Select previously disabled route and try to enable it again.
### Expected result
Checkbox should be unchecked and You should be able to save progress.
### Actual result
Exception will raise:
```
System.InvalidCastException: Unable to cast object of type 'BlendingEditor.Website.ViewModels.RoutingPage' to type 'BlendingEditor.Website.ViewModels.AllRoutesPage'.
   at lambda_method(Closure , RoutingItemPage )
   at lambda_method(Closure , Json , Input`1 )
   at Starcounter.Templates.Property`1.InvokeHandler(Json parent, Input`1 input) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter.XSON\Templates\Foundation\Property.cs:line 262
   at Starcounter.Templates.Property`1.ProcessInput(Json parent, T value) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter.XSON\Templates\Foundation\Property.cs:line 284
   at Starcounter.XSON.JsonPatch.DefaultPatchHandler.ParseAndProcess(TBool property, Json parent, String valueStr) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter.XSON\Sessions\JSON-Patch\DefaultPatchHandler.cs:line 74
   at Starcounter.Internal.TransactionManager.Scope(TransactionHandle handle, Action action) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter\TransactionManager.cs:line 552
   at Starcounter.Transaction.Scope(Action action) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter\Transaction.cs:line 225
   at Starcounter.Advanced.XSON.JsonExtension.Scope(Json json, Action action) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter.XSON\Advanced\JsonExtension.cs:line 69
[...]

Issue #29, opened 2018-01-09
Upgrade juicy-select to the current version


warpech:
Upgrade `juicy-select` with Bower to the current version https://github.com/Juicy/juicy-select/releases/tag/1.1.0

kegor:
Closing as done in PR #39.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Blending/releases/tag/5.2.0
- published in the App Warehouse

PR #34, opened 2018-01-10
Fixed Json merging when Db.Scope is used

un-tone:
For https://github.com/Starcounter/Blending/issues/6

I'm not completely sure what is the reason for the bug but can suppose that there is some problem with JSON merging when long-running transactions are used in view-models.


PatrykSzwer:
Since as You said there can be a bug with JSON merging, shouldn't You also replace this line: https://github.com/Starcounter/Blending/blob/7e8075d305b9af6aad80f0725fa4e7cb9ef907e2/src/BlendingProvider/Website/Api/ContentHandlers.cs#L219 ?

warpech:
Shouldn't this wait until https://github.com/Starcounter/Blending/issues/29 is done, considering the comments that you wrote in #6?

Or is this unrelated after all?

un-tone:
@warpech the bug is gone with this fix, so we should not wait for #29.

un-tone:
@PatrykSzwer, done.

warpech:
I like this, but I will test this manually before approving.

test/BlendingProvider.Tests/Test/DbScopeTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            WaitUntil(x => controlsPage.CancelButton.Enabled);
+            controlsPage.CancelButton.Click();
+            WaitUntil(x => !string.IsNullOrEmpty(controlsPage.CancelMessage.Text));
+
+            Assert.AreEqual("Cancel button was clicked 1 times", controlsPage.CancelMessage.Text);
+
+            var log = GetLog();
+            if (log != null)
+            {
+                var errors = Driver.Manage().Logs.GetLog(LogType.Browser).Where(x => x.Message.Contains("Starcounter.DbException: ScErrTransactionClosed"));
```
warpech: This seems too specific to me. It will pass if the error code is changed, even if the error occurs.

Can you change it from explicitly reading the console log to just perform one more action (like clicking the button for a second time)? If Palindrom does not work, that action should fail. Please double-check if this test fails before your fix and passes after your fix.

un-tone: Strange, I checked it without fixes. I'll double-check it, sure.

un-tone: And I'll replace this reading log

warpech: > Strange, I checked it without fixes. I'll double-check it, sure.

I just mean if you replace this, please check that again. Now I believe it is testing correctly.

un-tone: Removed that log checking. This one https://github.com/Starcounter/Blending/pull/34/files#diff-b566c5ed99c1d888ea36ea854781d3eeR54 is enough.

un-tone:
I've just found a bug with this fix when reproducing the case from  https://github.com/Starcounter/Blending/issues/31:

```
System.Exception: ScErrSessionJsonNotRoot (SCERR14010): The json is not a rootobject (have no parent) and cannot be used as sessiondata.
Version: 2.4.0.4481.
Help page: https://docs.starcounter.io/?q=SCERR14010.
   в Starcounter.Session.set_ClientRoot(Json value) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.XSON\Sessions\Session.cs:строка 633
   в Starcounter.XSON.JsonMimeConverter.Convert(Json json, MimeType mimeType, String& resultingMimeType) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.XSON\Class Json\JsonMimeConverter.cs:строка 52
   в Starcounter.Internal.ResourceMimeConverter.TryConvert(Request request, IResource resource, String mimeTypeStr, String& resultingMimeType) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.Apps.JsonPatch\BuiltInRestHandlers\ResourceMimeConverter.cs:строка 67
   в Starcounter.Internal.ResourceMimeConverter.Convert(Request request, Response response, String& resultingMimeType) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.Apps.JsonPatch\BuiltInRestHandlers\ResourceMimeConverter.cs:строка 46
   в Starcounter.Response.ConvertResource(Request req) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.Internal\Http\Response.cs:строка 771
   в Starcounter.Response.ConstructFromFields(Request req, Byte[] givenBuffer) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.Internal\Http\Response.cs:строка 827
   в Starcounter.Request.SendResponseInternal(Response resp, Byte[] serializationBuf) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.Internal\Http\Request.cs:строка 1195
   в Starcounter.Internal.AppsBootstrapper.ProcessExternalRequest(Request req) в C:\TeamCity\BuildAgent\work\sc-13028\Level1\src\Starcounter.Apps.JsonPatch\AppsBootstrapper.cs:строка 395
HResult=-2146233088
HelpLink=https://docs.starcounter.io/?q=SCERR14010
```

un-tone:
It seems that I found a solution - like it was done in https://github.com/Starcounter/Website/pull/300. The case works without errors, but now it requires to add a new test.

un-tone:
After the investigation, I found a better solution. Please check.

PatrykSzwer:
Looks like it works for me, but I guess We should wait for final review from @warpech

warpech:
I will review this in the afternoon.

un-tone:
There are some issues with tests - many of them have failed in SC 2.4.0.4627 (Jan 20) - on my local machine and on the TeamCity. I checked more than one time in both places (with wiping of local DB).

It looked really strange so I rebooted the Build Agent in TC. Also, I ran the tests without these changes - all tests passed. After it, I ran the tests in 2.4.0.4593 (Jan 18) - all worked well.

After rebooting of the Build Agent, all tests passed: https://teamcity.starcounter.org/viewLog.html?buildId=43874&tab=buildResultsDiv&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24 and https://teamcity.starcounter.org/viewLog.html?buildId=43876&tab=buildResultsDiv&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24.
Last time I ran tests on the local machine all work well too 😕. I think you should be aware of it.

cc @warpech @PatrykSzwer

un-tone:
Actually, it is not so good on the local machine with the tests. One time a test passes, another time it does not.
Just in case I prepared a revert PR https://github.com/Starcounter/Blending/pull/60 in advance.

warpech:

Let's continue the discussion about reverting or fixing the remaining issues in PR #60.

un-tone:
> released (with release notes) https://github.com/Starcounter/Blending/releases/tag/5.2.0

@warpech I guess it is not
image

warpech:
Oh, thanks! I will clean up my comment.

PR #36, opened 2018-01-10
Removed unnecessary code

un-tone:
Removed unnecessary `restoredFromSession` flag when getting a surface page.

This flag can be removed, since the condition which used it also checks `surfacePage.RequestUri != requestUri` and in case when the page was retrieved from Session, `surfacePage.RequestUri` is null and it makes the condition to return True. I ran all the tests - all passed. Also, I checked the app with different combinations of surfaces and routes.

The PR will be re-targeted to `master` after merging #34.


warpech:
This is a simplification of change that was made by @un-tone (with tests) in https://github.com/Starcounter/Website/pull/276, so I trust that it is correct!

warpech:
Please wait for PR #34 before merging

PR #38, opened 2018-01-11
Use all test projects in the TeamCity build

warpech:
I made a change to the TeamCity build script that allows to provide paths to multiple project DLLs, separated by a comma.

TeamCity was configured to include all paths to test DLLs and helpers. Let's see if the next build includes all tests.


warpech:
All 296 tests are passing!

warpech:

Issue #47, opened 2018-01-15
Palindrom-redirect stopped working


PatrykSzwer:
### Steps to reproduce
1. Go to `blendingeditor/surfacegroups`.
2. Create a new surface or select existing one.
3. Click `Delete` button inside general view.
### Expected result
User is redirected back to surfacegroups page.
### Actual result
Nothing is happening. That includes errors in console and exceptions.

PatrykSzwer:
Closing as done.

warpech:
Setting as “Finished” because it is:
- released (with release notes) https://github.com/Starcounter/Blending/releases/tag/5.2.0
- published in the App Warehouse

PR #48, opened 2018-01-15
Merge CssVariableManager to Blending app

PatrykSzwer:
I've migrated all needed views from CssVariableManager with all functionalities. I had to change couple of things, for the new views to be compatible, but main functionality should stay the same. 

Be aware that redirect from buttons in collection view don't work because of https://github.com/Starcounter/Blending/issues/47


src/BlendingEditor/wwwroot/BlendingEditor/views/CollectionPage.html
```diff
@@ -0,0 +1,131 @@
+<link rel="import" href="/sys/polymer/polymer.html">
+<link rel="import" href="/sys/juicy-color-picker/src/juicy-color-picker.html">
```
warpech: This file does not load (404).

src/BlendingEditor/CssVariablesManager/ViewModels/RenderedCollectionPage.json.cs
```diff
@@ -0,0 +1,41 @@
+using System.Text.RegularExpressions;
+using BlendingEditor.Website.Api.Authorization;
+using BlendingEditor.Website.Api.Authorization.Permissions;
+using Simplified.Ring6;
+using Starcounter;
+using Starcounter.Authorization.Attributes;
+
+namespace BlendingEditor.CssVariablesManager.ViewModels
+{
+    [PartialUrl("/blendingeditor/rendered-collections/{?}")]
```
warpech: Something is wrong here. Trying to access `http://localhost:8080/blendingeditor/rendered-collections/Eju` (a Render URI presented on the "Collection details" page) gives me 404.

src/BlendingEditor/wwwroot/BlendingEditor/views/CollectionPage.html
PatrykSzwer: For some reason I had to accidentally check it on CssVariablesManager itself, because I didn't get those errors before.

src/BlendingEditor/CssVariablesManager/ViewModels/RenderedCollectionPage.json.cs
PatrykSzwer: There should not be 404 anymore.

src/BlendingEditor/wwwroot/BlendingEditor/views/CollectionsPage.html
```diff
@@ -0,0 +1,34 @@
...
+                    <td class="blendingeditor-table__column--last">
+                        <button type="button" class="btn btn-sm btn-default" onclick="++this.value;" value="{{item.Delete$::click}}">Delete</button>
+                    </td>
+                </tr>
+            </template>
+            </tbody>
+        </table>
+        <button slot="blendingeditor/collections-add-collection" type="button" class="btn btn-primary" onclick="++this.value;" value="{{model.AddCollection$::click}}">Add collection</button>
+    </template></dom-bind>
+    <template is="starcounter-composition">
```
un-tone: Should be renamed to `declarative-shadow-dom`. It is the only such place in the project.

PatrykSzwer: Good catch! Didn't noticed that when I was migrating views.

PatrykSzwer:
I've tried to test it by adding blending rule to one of the blending points for surface, and trying to find new css variable in dev tools under surface preview, but couldn't find anything, same with this merged version and cssvariablesmanager itself, so I've probably done something wrong. @warpech could You please check it by urself?

PatrykSzwer:
I've moved rendered-collection to blendingProvider, please do the final check.

About issue that We spoke about, turns out that all We had to do is to place ID of an object in route URI inside curly braces.

warpech:
I am doing a review of this at the moment.

PatrykSzwer:
After #57 and #58 will be merged, I'll release a new version of Blending app with CssVariablesManager merge included.

warpech:

PR #56, opened 2018-01-17
Problem #53: Added warning dialog before deleting CSS Collections

### LINKED ISSUE #53: Ask user to confirm before deleting an item from CSS Variables
Moved from https://github.com/Starcounter/CssVariablesManager/issues/22

@Mihaiii wrote:

> Right now, deleting an item requires one click. A confirmation should be needed before an item is deleted as suggested [here](https://github.com/StarcounterApps/CssVariablesManager/pull/20#pullrequestreview-49479277) .

kegor:
**Solution**: Initial idea to fix #53 based on other code in project regarding using paper-dialog.

**Tests**: manual check.

**Screenshots**:
warning_dialog_before_delete_csscollection4


Mihaiii:
People app should serve as the official example, but Content app also have similar logic implemented and it's a simpler app so it could also be useful.

PatrykSzwer:
Yes, but as we speaking about Content app @Mihaiii, please take a look on that, and make sure It's running at SC 2.4, because right now I think It's not.

src/BlendingEditor/CssVariablesManager/ViewModels/CollectionsPage.json
```diff
@@ -4,8 +4,12 @@
     {
       "Name": "",
       "ViewUrl": "",
-      "Delete$": 0
+      "OpenCollectionsModalTrigger$": 0
```
PatrykSzwer: I would change the name on something more self-explanatory like `OpenDeleteCollectionModalTrigger`.

src/BlendingEditor/CssVariablesManager/ViewModels/CollectionsPage.json.cs
```diff
@@ -43,10 +57,11 @@ partial class CollectionsPageCollection : Json, IBound<CssCollection>
 
             protected CollectionsPage ParentPage => this.Parent.Parent as CollectionsPage;
 
-            void Handle(Input.Delete action)
+            void Handle(Input.OpenCollectionsModalTrigger action)
             {
-                this.ParentPage.Collections.Remove(this);
-                Db.Transact(() => this.Data.Delete());
+                this.ParentPage.selectedCollectionsPage = this;
+                this.ParentPage.ConfirmMessage = "Are you sure want to delete [" + this.Data.Name + "]?";
```
PatrykSzwer: Instead of entire message, lets just keep here only the selected collection name, and part that is always the same ("Are you sure..." etc) move to the view.

warpech:

PR #57, opened 2018-01-18
Palindrom-redirect fix

PatrykSzwer:
For #47 

I've used new implementation of palindrom-redirect tag, which fixed the problem. It should be done in any app that uses old link implementation and SC 2.4


src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -78,7 +78,7 @@ <h5>Are You sure You want to delete all compositions?</h5>
                 </template>
             </div>
         </paper-dialog>
-        <link is="palindrom-redirect" history url="{{model.RedirectUrl$}}">
+        <palindrom-redirect history url="{{model.RedirectUrl$}}" />
```
warpech: For HTML engine, this tag appears to be non-closed.

src/BlendingEditor/wwwroot/BlendingEditor/views/GeneralPage.html
```diff
@@ -43,6 +43,6 @@
             <a class="btn btn-xs btn-success" disabled="{{!model.Trn.IsDirty}}" value="{{model.Save$::click}}" onmousedown="++this.value;">Save</a>
             <a class="btn btn-xs btn-default" value="{{model.DeleteTrigger$::click}}" onmousedown="++this.value;">Delete</a>
         </div>
-        <link is="palindrom-redirect" history url="{{model.RedirectUrl$}}">
+        <palindrom-redirect history url="{{model.RedirectUrl$}}" />
```
warpech: For HTML5 engine, this tag appears to be non-closed. Replace this line with `<palindrom-redirect history url="{{model.RedirectUrl$}}"></palindrom-redirect>`

Finishing a tag declaration with `/>` does not mean anything for an HTML engine. HTML is not XML.

In HTML, there are several "void elements", that do not need a closing tag, but custom elements are not among them.

For more info, see:

- https://stackoverflow.com/questions/25313426/are-void-elements-and-empty-elements-the-same
- http://www.456bereastreet.com/archive/201005/void_empty_elements_and_self-closing_start_tags_in_html/

PatrykSzwer: Thanks for info, good to know! I'll fix that

warpech:

Issue #42, opened 2018-01-12
Adding new header rule for route is not working


PatrykSzwer:
### Steps to reproduce
1. Run Blending app and SignIn app.
2. Call `/blendingeditor/resetdata` 
3. Go to All routes tab.
4. Select any route and click add new header rule.
### Expected result
Header rule is added.
### Actual result
Exception:
```
System.InvalidOperationException: Manual operations on a bound array is not permitted. Array: MasterPage.AllRoutesPage.Routes.RoutingItemPage.Headers
   w Starcounter.Json.VerifyNotBound() w C:\TeamCity\BuildAgent\work\sc-13064\Level1\src\Starcounter.XSON\Class Json\Json.IList.cs:wiersz 483
   w Starcounter.Json.VerifyJsonForInserting(Nullable`1 index, Object item, Boolean skipVerification) w C:\TeamCity\BuildAgent\work\sc-13064\Level1\src\Starcounter.XSON\Class Json\Json.IList.cs:wiersz 397
   w Starcounter.Json.Add(Object item, Boolean skipVerification) w C:\TeamCity\BuildAgent\work\sc-13064\Level1\src\Starcounter.XSON\Class Json\Json.IList.cs:wiersz 172
   w BlendingEditor.Website.ViewModels.RoutingItemPage.Handle(AddHeaderTrigger action) w D:\
epositories\Blending\src\BlendingEditor\Website\ViewModels\RoutingItemPage.json.cs:wiersz 32
   w lambda_method(Closure , Json , Input`1 )
   w Starcounter.Templates.Property`1.InvokeHandler(Json parent, Input`1 input) w C:\TeamCity\BuildAgent\work\sc-13064\Level1\src\Starcounter.XSON\Templates\Foundation\Property.cs:wiersz 262
   w Starcounter.Templates.Property`1.ProcessInput(Json parent, T value) w C:\TeamCity\BuildAgent\work\sc-13064\Level1\src\Starcounter.XSON\Templates\Foundation\Property.cs:wiersz 284
   w Starcounter.XSON.JsonPatch.DefaultPatchHandler.ParseAndProcess(TLong property, Json parent, String valueStr) w C:\TeamCity\BuildAgent\work\sc-13064\Level1\src\Starcounter.XSON\Sessions\JSON-Patch\DefaultPatchHandler.cs:wiersz 149
[...]

PR #58, opened 2018-01-18
Fix for adding new header rules for routes

PatrykSzwer:
For #42


warpech:

PR #60, opened 2018-01-20
Revert "Fixed Json merging when Db.Scope is used"

un-tone:
Reverts Starcounter/Blending#34 - look there.


warpech:
This is a priority. Let's investigate if the revert is needed. 

It is much preferred to find the reason why the tests pass only sometimes.

un-tone:
I think this can be closed.

Issue #65, opened 2018-02-02
Apply the new jargon


warpech:
Since we have now agreed on the new jargon (https://github.com/Starcounter/RebelsLounge/issues/227#issuecomment-361909655), the following names in BlendingEditor should be replaced:

- Surface → Layout
- Routes → Layout conditions
- Blending point → Area
- Blending rule → Area items

@kegor could you please do that in:

- UI
- app logic (view-models, etc)
- URIs
- README.md

-----

I suggest to base the work on https://github.com/Starcounter/Blending/pull/62 branch.

kegor:
Closed because (I hope) mostly the all renamings were performed and merged in #66.

PR #66, opened 2018-02-04
Problem #65: Apply the new jargon

kegor:
Initial iteration for #65.

Working on fixing tests right now.


kegor:
For now, only tests that already mentioned in #67 are failed.

src/BlendingEditor/Website/ViewModels/BaseLayoutConditionsPage.json.cs
```diff
@@ -0,0 +1,55 @@
...
+            this.ModalOpened = true;
+        }
+
+        public void Handle(Input.SaveChangesTrigger action)
+        {
+            foreach (var layoutCondition in this.LayoutConditions)
+            {
+                layoutCondition.HeadersString = layoutCondition.Headers.Select(x => x.Data).GetHeadersString();
+                var emptyHeaders = layoutCondition.Headers.Where(h => string.IsNullOrWhiteSpace(h.Name)).ToList();
+                foreach (var routeHeader in emptyHeaders)
```
PatrykSzwer: there are still some occurrences of `routeHeader`. Please fix here and in rest places.

src/BlendingProvider/Website/Api/BlendingHandlers.cs
```diff
@@ -85,12 +85,12 @@ public void Register()
         /// <summary>
         /// Map Blending Rule
```
PatrykSzwer: There are still some comments that refers to `blending rules` which should be changed to are items. Please change them.

kegor: Done.

src/BlendingEditor/Website/ViewModels/BaseLayoutConditionsPage.json.cs
kegor: Done.

warpech:
Please merge current `master` to this branch. It should succeed now.

warpech:
In https://github.com/Starcounter/Blending/pull/66/commits/6d648a2e93d67a88e99071f0e217133b9a2eaaca, I fixed few instances where there was some error in translation (the term"area item" was used for header rules of layout conditions). @PatrykSzwer could you please double check if there are no other mistranslations like that?

Issue #67, opened 2018-02-05
The tests do not pass with Starcounter newer than 2.4.0.4692


warpech:
I discovered this with the branch of PR https://github.com/Starcounter/Blending/pull/62 (`55-add-automatic-tests`), but I suppose that the `master` branch is equally affected.

The app is tested successfully on TeamCity with Starcounter 2.4.0.4692. Any version of Statcounter newer than that causes from 4 to 6 tests to fail:

```	
WebStuffDisablingTests(Chrome,DisableViewKeeper).ReenableDuplicatedBlendingRule_ThisRuleContentNotLoaded  	
WebStuffDisablingTests(Chrome,EnableViewKeeper).ReenableDuplicatedBlendingRule_ThisRuleContentNotLoaded  	
WebStuffDisablingTests(Firefox,DisableViewKeeper).ReenableDuplicatedBlendingRule_ThisRuleContentNotLoaded  	
WebStuffDisablingTests(Firefox,EnableViewKeeper).ReenableDuplicatedBlendingRule_ThisRuleContentNotLoaded 
```

@PatrykSzwer, @kegor please investigate what might be the reason?

PatrykSzwer:
It seems like some one-version issue, because either locally and teamcity there are no more failing tests.

kegor:
Closed because the fix was merged in #68.

warpech:
Is there any good explanation why did it start to fail only recently?

PatrykSzwer:
We talked a little with Anton about it and We came up with only 2 possibilities: 
1. It can be a coincidence that It appeared just now.
2. There had to be some low-level change in obtaining data that triggered this issue to bubble up.

PR #68, opened 2018-02-06
Problem #67: The tests do not pass with Starcounter newer than 2.4.0.4692.

kegor:
Possible fix for #67.


warpech:
Is there any good explanation why did it start to fail only recently?

Issue #75, opened 2018-02-12
Make a page about composition editor


warpech:
As discussed on the today's meeting, add a new page "Composition Editor" to the menu.

This page should have the following static content for the beginning:

> # COMPOSITION EDITOR
>
> Composition editor is a feature of BlendingEditor allowing you to create and manage custom compositions of partial HTML views coming from blended apps.
>
> To open the composition editor, press `CTRL`+`E` on any page. The editor will open on the top of the screen.
>
> For more information about using the composition editor, see the docs: https://docs.starcounter.io/guides/web-apps/view-composing

Is this enough for the beginning? In future we can make changes to this page as separate PRs.

@kegor please start the work from the PR #63 and add @PatrykSzwer as the reviewer.

kegor:
Closed because the fix was merged in #78.

PR #78, opened 2018-02-13
Problem #75: Make a page about composition editor

kegor:
**Task #75:**
Page with static content about Composition Editor should be added into Blending Editor menu.

**Solution:**
Page was added.

**Screenshots:**
image


src/BlendingEditor/wwwroot/BlendingEditor/views/CompositionEditorPage.html
```diff
@@ -0,0 +1,12 @@
+﻿<template>
+    <h2>Composition Editor</h2>
+    <p>
+        Composition editor is a feature of BlendingEditor allowing you to create and manage custom compositions of partial HTML views coming from blended apps.
+    </p>
+    <p>
+        To open the composition editor, press CTRL+E on any page. The editor will open on the top of the screen.
```
PatrykSzwer: I would bold CTRL+E part, for better readability.

src/BlendingEditor/wwwroot/BlendingEditor/views/CompositionEditorPage.html
```diff
@@ -0,0 +1,12 @@
+﻿<template>
```
PatrykSzwer: Since we want to have all views in blending app with declarative-shadow-dom, please add slots etc. to this view too, for future reference.

src/BlendingEditor/wwwroot/BlendingEditor/views/CompositionEditorPage.html
kegor: Done.

src/BlendingEditor/wwwroot/BlendingEditor/views/CompositionEditorPage.html
kegor: On the CollectionsPage I saw the slot for <h2> and I added here the same, but I'm not sure about the slots for other <p>-tags. So I'm rather looking forward to your suggestions on this matter.

warpech: See https://docs.starcounter.io/guides/web-apps/html-view-guidelines#guideline-3:-attaching-the-content-to-slots:

>  Use explicit slot names instead of relying on the default slot.

kegor: That's done, if I understood correctly.

src/BlendingEditor/wwwroot/BlendingEditor/views/CompositionEditorPage.html
```diff
@@ -0,0 +1,22 @@
...
+    <dom-bind>
+        <template is="dom-bind">
+            <h2 slot="blendingeditor/compositioneditor-heading">Composition Editor</h2>
+            <div slot="blendingeditor/compositioneditor-description">
+                <p>
```
PatrykSzwer: It's still incorrect. You should move div that spans all those paragraphs to declarative-shadow-dom and put slots on all p elements (You can use same slot name for all of them)

kegor: I guess I eventually got the idea (after re-reading the [docs](https://docs.starcounter.io/guides/web-apps/html-view-guidelines#guideline-4:-create-the-layout-in-declarative-shadow-dom)), and also I decided to remove `<div>` because for me it looks redundant for the moment.

Issue #85, opened 2018-02-20
Highlight the current link in the menu using a custom element in Shadow DOM


warpech:
Leftover from https://github.com/Starcounter/Blending/pull/63#issuecomment-366098556

warpech:
Implemented in https://github.com/Starcounter/Blending/pull/90

warpech:
After SCF 3.2.0 is shipped with Starcounter, remove `<highlight-current-link>` from Blending as it is no longer needed (now the equivalent is provided by the Starcounter app shell)

warpech:
> After SCF 3.2.0 is shipped with Starcounter, remove `<highlight-current-link>` from Blending as it is no longer needed (now the equivalent is provided by the Starcounter app shell)

This is done.

----

Left to do:

- add information on the https://docs.starcounter.io/v/2.4/topic-guides/blendable-web-apps/app-shell page that App shell automatically adds `aria-current="page"` (or is the value `"true"`?) to the current links. Apps can leverage this information, for example Uniform.css `XXXXX` class uses it to highlight the current link in the menu.
- add information in KitchenSink's WelcomePage saying that the links in the menu are highighted thanks to the Starcounter app shell with a link to https://docs.starcounter.io/v/2.4/topic-guides/blendable-web-apps/app-shell

@PatrykSzwer will do that.

warpech:
Done in the two PRs linked above.

PR #90, opened 2018-02-22
implement highlight-current-link (#85)

warpech:
Explained in the commit message.

@PatrykSzwer pls review. This should be merged before doing https://github.com/Starcounter/uniform.css/issues/44


src/BlendingEditor/wwwroot/BlendingEditor/views/MasterPage.html
```diff
@@ -24,6 +24,59 @@
     </template></dom-bind>
 
     <template is="declarative-shadow-dom">
+        <script>
+        class HighlightCurrentLink extends HTMLElement {
+            connectedCallback() {
+                /*
+                does not work on page load in polyfilled browsers
```
PatrykSzwer: Is this comment necessary? If not, please remove it.

src/BlendingEditor/wwwroot/BlendingEditor/views/MasterPage.html
```diff
@@ -24,6 +24,59 @@
     </template></dom-bind>
 
     <template is="declarative-shadow-dom">
+        <script>
```
PatrykSzwer: Could you move this definition of custom element to some separate file and import it here? I think masterPage will be more readable that way.

warpech: As said today on the status meeting, I would like to discuss outside of this PR about where to put such custom elements. I have made an issue: https://github.com/Starcounter/RebelsLounge/issues/246

warpech: In case of this element, we want all apps to be able to use it. So it makes little sense to include it as an explicit HTML Import in BlendingEditor, because other apps will have to duplicate it.

PatrykSzwer: I didn't mean this moving to file in terms of usage, more in case of readability.

Issue #10, opened 2018-01-08
🤸 Select Blend URI option


PatrykSzwer:
_Issue moved from Website repository: https://github.com/Starcounter/Website/issues/234_

In Content app I use the id for URIs (ex: `/content/dynamic/Cq`).

In version 1 of Content app, the user had the possibility to write the URIs that he/she desires (after `/content/dynamic/`), but I changed this because now a Content Item can have a `Name` and there can be multiple Content Items with the same name (especially if the parent content item is different. For example, "Introduction" can be a content item name that is repeating).

So at this moment the user has to select the URI that is provided when he/she creates a new content item and copy it to "Blend URI". I don't find this very friendly.

I was thinking about having a combo box for Blend URI, maybe? (with displays like `/content/dynamic/Cq ({ContentItemName})` if the values are from `Simplified.Ring6.WebContent`)
What do you think?

warpech:
@kegor could you please check if there is an API that allows us to get a list of currently declared URIs? It should be the same list that is used to populate http://localhost:8181/#/server/network. If you can't find it, please ask @alemoi.

kegor:
@warpech you mean something like http://127.0.0.1:8181/gw/stats?

warpech:
Yes, but you want a C# method that returns a list of those, not a REST API call.

kegor:
@warpech I found one example here https://github.com/Starcounter/level1/blob/2ee996fbbd7c5201d2b82d3c8c7493569ab237db/src/Programs/Starcounter.TrayIcon/Service/NetworkTask.cs#L21.

warpech:
I was rather thinking of [GetRegisteredHandlers](https://github.com/Starcounter/level1/blob/cb13897cb2352deb726d91e2463340e28b45d234/src/Starcounter.Internal/Handle.cs#L470) which seems to be publicly exposed, but not documented. @alemoi can we use that?

alemoi:
> @alemoi can we use that?

Sure but it will give you handlers for current codehost/database, not all.

Mihaiii:
Content app is blocking the original issue that was opened by me in Website app.
I really appreciate what @warpech did (I'm not sure if it was intentional or as a side effect) and I'd like to announce that Content app will be fixed by monday for SC 2.4, in case this will change anything regarding this current issue.

warpech:
> Sure but it will give you handlers for current codehost/database, not all.

That's completely understandable, thanks. It is enough.

[...]

PR #93, opened 2018-02-24
Problem #10: Select Blend URI option

kegor:
**Task #10:**
We are thinking about having a combo box for Blend URI which looks more user-friendly than text input field.

**Solution:**
In the initial iteration was added vaadin-combo-box that uses results of Handle.GetRegisteredHandlers as values.

**Tests:**
There are no tests provided for the moment.

**Screenshots:**
blend_uris_vaadin_combo_box_wide


src/BlendingEditor/Website/Helpers/HandleHelper.cs
```diff
@@ -0,0 +1,13 @@
+﻿using Starcounter;
+using Starcounter.Internal;
+
+namespace BlendingEditor.Website.Helpers
```
kegor: Maybe I could move in this class all contents from https://github.com/Starcounter/Blending/blob/master/src/BlendingEditor/Website/Helpers/HandleFile.cs?

PatrykSzwer: It's structure that came from kitchensink and it can be useful that those HandleFile functions are in a separated file. I would leave this as it is right now.

kegor:
Yesterday I had an interesting discussion with @Mihaiii, we were talking about this view of the vaadin-combo-box in People app:
styled_vaadin_combo_box
Mihai pointed that the version of vaadin-combo-box in People app is older (2.0.0) and this why more simple view of this component in Blending app is not the issue. I checked version and now I see that in current PR Blending app got 3.0.3 version of vaadin-combo-box.

_Interesting fact:_
If I make run for People app I see old version (2.0.0) of vaadin-combo-box (that looks more styled) in both apps and if I make run for Blending app I see in both apps new (3.0.3) version of vaadin-combo-box that looks more simple.

kegor:
@PatrykSzwer I'd increased the width of the vaadin-combo-box with a small reserve.

PatrykSzwer:
There is some strange behavior that I'm not sure what to think about.
When you add really long URI, combobox will have 3(!) sliders:
image
As far as I'm okay with two of them, the horizontal slider doesn't show for me all the time, which is not good since I would like to have a possibility to check my entire really long URI. On the other hand, I think that those 2 vertical sliders look odd but TBH I'm not sure what can be done for that matter.

PatrykSzwer:
Please try to use 4.0.0 beta version of vaadin-combo-box.

kegor:
This is the new view of our URI's picker with vaadin-combo-box#4.0.0-beta3:
400_beta_vaadin_combo_box
@PatrykSzwer, what do you think about this new style?

warpech:
Fiddling with the demo page https://vaadin.com/elements/vaadin-combo-box/html-examples/combo-box-basic-demos, I have noticed that you can apply custom height to this input by setting a variable:

```
--lumo-size-m: 38px !important;
```

Please find what is the desired exact amount in pixels and set that value in `declarative-shadow-dom` of our view.

PatrykSzwer:
Also please fix conflicts with `master` branch.

warpech:
I suggested to @kegor on Slack to actually keep the default `--lumo-size-m`. The solution designer can overwrite that value with CSS variables manager.

PatrykSzwer:
There is some test that failed, please investigate.

kegor:
@PatrykSzwer seems like this was `flaky` fail for this test. Next build was successful.

PatrykSzwer:
I would still suggest to set this variable some new value, because with default one it just looks ugly.

kegor:
I found that vaadin-combo-box looks ok if I removing content of Underwear.css:
image
and @warpech prompted that problem is in this line https://github.com/Starcounter/underwear.css/blob/c7a007ce188130e4bb0c2edcf29bff959912416b/src/underwear.css#L114.
So we should change this from `font-size: 62.5%` to `var(--default-font-size)` in another Underwear.css bug.

kegor:
Bug for Underwear.css was created: https://github.com/Starcounter/underwear.css/issues/68.

Issue #88, opened 2018-02-21
🤸 Select CSS variable collections to be provided with a layout


warpech:
Currently, it is possible to define CSS variable collections in BlendingEditor, but to use them in a layout, you need to explicitly put the URI of the collection as an area item.

## Selecting the collections

To make this easier, we should add a possibility to select the CSS variable collection on the "General" page of a layout.

I imagine it like this:

> ![image](https://user-images.githubusercontent.com/566463/36494084-698425a6-1731-11e8-93cb-8c4d40eb80c7.png)

When you press the <kbd>Select collections</kbd> button, a modal opens with a checkbox containing the list of available CSS variable collections. You can select more than one.

## Providing the collections

For all selected collections, BlendingProvider should include that collection in the light DOM part of [layout view HTML handler](https://github.com/Starcounter/Blending/blob/6daec49cfaf6e10bd21e7451bb71a629ae4f264d/src/BlendingProvider/Website/Api/ContentHandlers.cs#L63-L89)

@PatrykSzwer, @kegor WDYT?

kegor:
In general, this sounds more UX friendly.

PatrykSzwer:
I definitely agree. Since we have CssVariablesManager merged to blending it should be relatively easy to do.

warpech:
Great! @PatrykSzwer can you pls work on that next week then?

PatrykSzwer:
Closing as done.

PR #96, opened 2018-02-27
Layout variables collections

PatrykSzwer:
For #88 

Be aware that this PR needs those changes: https://github.com/StarcounterApps/Simplified/pull/108 to work.


src/BlendingProvider/Website/Api/ContentHandlers.cs
```diff
@@ -71,7 +71,27 @@ public void Register()
                     html += $"<starcounter-include slot=\"blendingeditor/{name}\" view-model=\"{{{{model.Sections.{name}}}}}\"></starcounter-include>\n";
                 }
 
-                html += "</template></dom-bind>\n<template is=\"declarative-shadow-dom\">";
+                // Part to add included css collections variables
+                html += "<style>";
+                foreach (var name in areaNames)
+                {
+                    html += $"starcounter-include[slot=\"blendingeditor/{name}\"]";
+                    if (areaNames.IndexOf(name) != areaNames.Count - 1) //For not adding ',' in last selector
```
kegor: Maybe this could be simplified like: `html = String.Join(",\n", areaNames.Select(name => string.Format($"starcounter-include[slot=\"blendingeditor/{name}\"]", name)));`.

PatrykSzwer: Good suggestion with this string.Join, I was looking for this solution but completely forgot about this method. Although this string.Format is unnecessary, string interpolation replaced string.format.

src/BlendingProvider/Website/Helpers/DataHelper.cs
```diff
@@ -1,11 +1,31 @@
...
 {
     public static class DataHelper
     {
+        public static string GetComputedVariables(IEnumerable<CssVariableValue> variables)
+        {
+            Regex areaItem = new Regex("^[a-zA-Z0-9#()\\s.,\\-!:_/\"\']+$"); //TODO change regexp to escaping of illegal characters
+            string variablesString = string.Empty;
+            foreach (CssVariableValue val in variables)
+            {
+                if (val.CssVariable.Name != null && areaItem.IsMatch(val.CssVariable.Name))
```
kegor: Maybe this `if`'s could be combined? (as described [here](https://en.wikibooks.org/wiki/Computer_Programming/Coding_Style/Minimize_nesting#Solutions)).

PatrykSzwer: TBH I just moved this method from CssVariablesManager old implementation, I didn't rewrite it, but sure, why not.

kegor:
Seems working:
image
Maybe table in the pop-up window looks not very stylish, and maybe the sequence of layout collections might be separated by a comma, but this is only for my non-designer taste.

src/BlendingEditor/wwwroot/BlendingEditor/views/GeneralPage.html
```diff
@@ -12,16 +13,28 @@ <h3 slot="blendingeditor/general-header">General</h3>
...
+            <label slot="blendingeditor/general-collections-label">This layout includes no collections.</label>
+        </template>
+        <template is="dom-if" if="{{model.LayoutCollections.length}}">
+            <label slot="blendingeditor/general-collections-label">This layout includes the following collections:
+                <template is="dom-repeat" items="{{model.LayoutCollections}}">
+                    <a href="/blendingeditor/collections/{{item.Key}}">{{item.Name}}</a>
+                </template>
+            </label>
+        </template>
+        <button slot="blendingeditor/general-collections-button" value="{{model.OpenModalTrigger$::click}}" onmousedown="++this.value;">Select collections</button>
```
warpech: image

This UI is non-intuitive without the header "CSS variable collections".

src/BlendingEditor/wwwroot/BlendingEditor/views/GeneralPage.html
```diff
@@ -34,6 +47,39 @@ <h3 slot="blendingeditor/general-preview-header">Preview</h3>
...
+                                <td>
+                                    <label>{{item.Name}}</label>
+                                </td>
+                                <td style="align-items: center">
+                                    <input type="checkbox" checked="{{item.Include$::change}}">
+                                </td>
+                            </tr>
+                        </template>
+                        </tbody>
+                    </table>
```
warpech: When no CSS variable collections are defined, I get an empty table:

image

It would be great to get some information that before I can select a CSS variable collection, I need to create one first.

src/BlendingProvider/Website/Api/ContentHandlers.cs
```diff
@@ -71,7 +71,20 @@ public void Register()
...
+                string variables = string.Empty;
+                var collections = Db.SQL<CssCollection>($"SELECT ccr.{nameof(CssCollectionRelation.Collection)} FROM {typeof(CssCollectionRelation)} ccr " +
+                                                        $"WHERE ccr.{nameof(CssCollectionRelation.Template)}.Key = ?", layoutId);
+
+                foreach (var collection in collections)
+                {
+                    variables += DataHelper.GetComputedVariables(collection.CssVariableValues);
+                }
+
+                html += $"{{{variables}}}";
```
warpech: This approach seems to work well for Underwear variables, but not for Uniform variables. 

Look at the below screenshot. I would expect:

- all whites to be yellow ✅
- "Save" button to be purple ❌

Maybe add `!important` to line 83?

Please check the solution in Chrome and Firefox.


image

src/BlendingProvider/Website/Api/ContentHandlers.cs
```diff
@@ -71,7 +71,20 @@ public void Register()
                     html += $"<starcounter-include slot=\"blendingeditor/{name}\" view-model=\"{{{{model.Sections.{name}}}}}\"></starcounter-include>\n";
                 }
 
-                html += "</template></dom-bind>\n<template is=\"declarative-shadow-dom\">";
+                // Part to add included css collections variables
+                html += "<style>" + string.Join(",\n", areaNames.Select(name => $"starcounter-include[slot=\"blendingeditor/{name}\"]"));
```
warpech: Unfortunately, this approach requires to refresh the window to see the changes. If you used the approach of exposing the style as a `CssVariableCollectionsRendering` in the view-model and importing it as `<style inner-h-t-m-l="{{model.CssVariableCollectionsRendering}}"`, we would see the changes live. That would be awesome.

src/BlendingEditor/wwwroot/BlendingEditor/views/GeneralPage.html
warpech: If you looked at it for the first time, would you know what collections do we mean here? I don't think so... I think there must be a header.

src/BlendingEditor/wwwroot/BlendingEditor/views/GeneralPage.html
warpech: > It would be great to get some information that before I can select a CSS variable collection, I need to create one first.

Maybe also add a link? Like:

[Manage your CSS variable collections](/blendingeditor/link)

PatrykSzwer:
There is one more test that is failing but It's not related to this branch (issue: https://github.com/Starcounter/Blending/issues/106) 
@warpech please confirm that I can merge this PR.

PatrykSzwer:
You can find created issue here: https://github.com/Starcounter/Blending/issues/108

Issue #30, opened 2018-01-10
⏰JsonPatchException: Unsupported patch operation in patch when saving composition


PatrykSzwer:
### Steps to reproduce
1. Run Blending app and People app.
2. do `/blendingeditor/resetdata`
3. Go to `/people/persons`
4. Turn on compositionEditor (Ctrl + E)
5. Select `[slot="blendingeditor/Main"]` from editor dropdown.
6. Change anything in the composition (e.g add some text on the end)
7. Click save.
### Expected result
No errors, composition is saved.
### Actual result
Exception:
```
Starcounter.XSON.JsonPatch.JsonPatchException: Unsupported patch operation in patch: 'Add'. Patch: '{"op":"add","path":"/BlendingProvider_0/Sections/Main/BlendingProvider_0/Composition$","value":"<style>:host{display:block}</style>\
        <slot name=\"BlendingProvider/SurfaceCachedActive\"></slot>\
    <slot></slot>ss"}'. Viewmodel versions: {client (_ver#c$): 1, clients serverversion: 0, server (_ver#s): 0}.
   at Starcounter.XSON.JsonPatch.DefaultPatchHandler.Handle(Json root, JsonPatchOperation patchOp, JsonPointer pointer, String value) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter.XSON\Sessions\JSON-Patch\DefaultPatchHandler.cs:line 17
   at Starcounter.XSON.JsonPatch.JsonPatch.HandleOnePatch(Json root, JsonPatchOperation op, JsonPointer ptr, String value, Int32 patchCount, ChangeLog changeLog, Object sourceJson, Boolean strictRejection, JsonPatchStatus previousStatus) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter.XSON\Sessions\JSON-Patch\JsonPatch.cs:line 603
   at Starcounter.XSON.JsonPatch.JsonPatch.Apply(Json root, JsonTextReader reader, Boolean strictPatchRejection, Object sourceJson, Int32& patchCount) in C:\TeamCity\BuildAgent\work\sc-12991\Level1\src\Starcounter.XSON\Sessions\JSON-Patch\JsonPatch.cs:line 524
[...]

PR #97, opened 2018-02-27
Problem #30: JsonPatchException: Unsupported patch operation in patch when saving composition

kegor:
Partial fix for #30.


tomalec:
I expressed my concerns at https://github.com/Starcounter/Blending/issues/30#issuecomment-369183983

warpech:
Let's freeze that until https://github.com/Starcounter/Blending/issues/30#issuecomment-369301653 is answered.

warpech:
We did not agree on this for a long time. Even if this PR is good, it is now obsolete because we are now targetting the mapping version of Starcounter. If needed, it will be easy to recreate this PR that consisted of just one commit: 3a3ef0bf0a261341817f84becccc3a7eebfccfb0

PR #98, opened 2018-02-27
Fix typos and grammar errors in texts

warpech:
As in the topic


src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -20,14 +20,14 @@
...
         <label slot="blendingeditor/settings-export-note">
-            Export all configuration data from current database to given file lokalization.<br> Exported will be: Layout groups, layouts, areas, area items, layout conditions and compositions.
+            Export all configuration data from the current database to a database dump file.<br> The following data will be exported: Layout groups, layouts, areas, area items, layout conditions and compositions.
         </label>
         <button slot="blendingeditor/settings-export-button" disabled="{{model.ExportInProgress}}" onmousedown="++this.value;" value="{{model.ExportDataTrigger$::click}}">{{model.ExportButtonText}}</button>
         <template is="dom-if" if="{{model.ExportInProgress}}">
             <div slot="blendingeditor/settings-export-loader" class="blendingeditor-loader"></div>
         </template>
 
-        <label slot="blendingeditor/settings-import-note">Import configuration data from select file for current database.</label>
```
kegor: If the word "selected" does not have a critical meaning, then everything looks like 5 out of 5.

warpech:
In case of a flaky test, we should re-run the tests on TC until it gets green.

I just did that for this PR.

I agree that we should fix the flaky tests, but it seems that they fail to seldomly for TC to detect them. I think it is better to wait until they appear on the page: https://teamcity.starcounter.org/project.html?projectId=Apps_Blending&tab=flakyTests&branch_Apps_Blending=%3Cdefault%3E

PR #102, opened 2018-02-28
🤸 Shorten FOUC, by using `link` instead of `@import` and `preload`

tomalec:
https://github.com/Starcounter/RebelsLounge/issues/244


warpech:
TL;DR replace `<style>@import url "..."</style>` with both:

- `<link rel="preload" href"...">` before the first `<template>`
- `<link rel="stylesheet" href"...">` in shadow DOM

tomalec:
Merge conflicts resolved

src/BlendingEditor/wwwroot/BlendingEditor/views/AreasPage.html
```diff
@@ -1,4 +1,6 @@
 ﻿<link rel="import" href="/sys/juicy-select/juicy-select.html">
+<link rel="preload" type="stylesheet" href="/sys/uniform.css/uniform.css">
```
warpech: Should be `link rel="preload" as="style"`, according to https://www.smashingmagazine.com/2016/02/preload-what-is-it-good-for/

tomalec: Fixed

warpech:
One question though: is it possible to keep `link rel="preload"` in Shadow DOM?

tomalec:
> is it possible to keep `link rel="preload"` in Shadow DOM?

What would it give? It will start downloading exactly same moment when regular `link`

warpech:
> What would it give? It will start downloading exactly same moment when regular link

1. It would still dedupe the network request (`link rel="stylesheet"` does not dedupe the requests)
2. As the solution designer, you would be able to remove preloading of Uniform.css when you decide to remove Uniform classes in a custom composition

tomalec:
> 1. It would still dedupe the network request (link rel="stylesheet" does not dedupe the requests)

But you will still need `<link rel="stylesheet">`. Please provide the code of what you mean in https://github.com/Starcounter/Blending/pull/102#issuecomment-369274025

warpech:
Of course! As I wrote:

> is it possible to keep `link rel="preload"` in Shadow DOM? It makes sense to preload only when we know that the stylesheet is actually used.

I mean to do this:

```html
<template>
    foo
        <link rel="preload" as="style" href="/sys/uniform.css/uniform.css">
        <link rel="preload" as="style" href="/BlendingEditor/css/blendinguniform.css">
        <link rel="stylesheet" href="/sys/uniform.css/uniform.css">
        <link rel="stylesheet" href="/BlendingEditor/css/blendinguniform.css">
        bar
    </template>
</template>
```

Instead of this:

```html
<link rel="preload" as="style" href="/sys/uniform.css/uniform.css">
<link rel="preload" as="style" href="/BlendingEditor/css/blendinguniform.css">
﻿<template>
    foo
        <link rel="stylesheet" href="/sys/uniform.css/uniform.css">
        <link rel="stylesheet" href="/BlendingEditor/css/blendinguniform.css">
        bar
    </template>
</template>
```

tomalec:
Then what is the benefit of keeping `<link rel="preload" as="style" href="/sys/uniform.css/uniform.css">` just before `<link rel="stylesheet" href="/sys/uniform.css/uniform.css">`?  AFAIK it will behave just like having `<link rel="stylesheet" href="/sys/uniform.css/uniform.css">` as we had before.

warpech:
I wrote two benefits https://github.com/Starcounter/Blending/pull/102#issuecomment-369586119, are they wrong?

tomalec:
1. Will not preload a thing, as you preload it exactly at the time you are loading it. and I don't think it will dedupe it either.

tomalec:
However, it may be valid point that preloading it before the template is stamped, is not needed at all.

PR #104, opened 2018-03-01
Provisional implementation of uni-title 

warpech:
The intended usage of `.uni-title` is to put a commonly formatted header inside `uni-section` or `uni-card`. It adjusts margins and forces the font size to 20px, even if the slotted element had a different font size.

It is an implementation of https://github.com/Starcounter/uniform.css/issues/43#issuecomment-368957379

If accepted, the definitions from `blendinguniform.css` have to be moved to `uniform.css`.

----

Other changes:

- changed the elements that are used in content for page title to `<h2>` (it doesn't matter actually)
- changed the elements that are used in content for card title to `<h3>` (it doesn't matter actually)


PatrykSzwer:
I've reviewed the code and run app itself. One comment though. Since with this class, It doesn't matter what element we use to provide header text, maybe we should just use `label` instead of messing around with different headers? Since here we're not using headers properties anyway.

warpech:
In my opinion, the `<hx>` headers still have semantic information for SEO, screen readers, etc. They have stronger meaning that a label of a form element. They are just displayed in a fixed font size when wrapped in `uni-title`.

warpech:
Thanks. It will be updated to use new Uniform.css in a separate PR next week.

I will fix the 4 failing tests and merge.

PR #105, opened 2018-03-04
Provisional implementation of uni-layout-with-left-side-nav

warpech:
Proposed explanation for Uniform.css demo page:

> ## Layout
>
> ### Layout with left side navigation
>
> Class `uni-layout-with-left-side-nav` implements a commonly used two-column layout.
>
> It uses two child elements:
>
> - first element child is the left column used for navigation
> - second element is the right column used for the main content
>
> Anchor elements (`<a>`) element within the left column that have the attribute `aria-current="true" will be highlighted.

-----

After this PR is accepted, I will move the code to Uniform.css. 

Also, after the other PR https://github.com/Starcounter/StarcounterClientFiles/pull/55 is accepted, this view can be greatly simplified.


src/BlendingEditor/wwwroot/BlendingEditor/views/MasterPage.html
```diff
@@ -76,78 +76,59 @@
...
 
-            .blendingeditor-nav-menu-panel a:hover {
-                color: white !important;
+            .uni-layout-with-left-side-nav > :first-child a:hover {
+                color: var(--uni-white) !important;
             }
 
-            .blendingeditor-nav-menu-panel ::slotted(a[aria-current="page"]) {
+            .uni-layout-with-left-side-nav > :first-child ::slotted(a[aria-current="page"]) {
                 border-left: 3px solid #0184b4 !important;
```
PatrykSzwer: Can You replace this hardcoded color with one from uniform css variables, like with the rest of them? Like `--uni-blue` or `--uni-dark-blue`.

warpech: I didn't do that because such exact color does not exist in Uniform.css. 

I suggest to merge it as is and do a PR in Uniform.css. If you find a close-enough color in Uniform.css, please use it instead of `#0184b4`. Otherwise add a new variable there.

warpech:
> One comment about the code. I also suggest simplifying the name of this class at least a little because It's too long right now. My first suggestion is to get rid of `with` and `side` words because those are not needed to get the purpose of the class.

I liked the verbosity of `.uni-layout-with-left-side-nav`, but I can also agree to `.uni-layout-left-nav`.

@alshakero any other suggestions perhaps?

alshakero:
I lean to shorter names more than verbose ones, because I usually have to memorize the name anyways. Not its meaning to me. I don't know about others though.

I suggest `.uni-layout-side-nav`, `side-nav` phrase is widely-adopted see [this](https://www.w3schools.com/howto/howto_js_sidenav.asp) and [this](http://materializecss.com/side-nav.html) and [that](https://material.angularjs.org/latest/demo/sidenav).

warpech:
In https://material.angularjs.org/latest/demo/sidenav, the class is called `md-sidenav-left`.

In http://materializecss.com/side-nav.html, `side-nav` does not contain the word `left` or `right` because it is undetermined and provided as part of the config (`edge: 'right', // Choose the horizontal origin`).

In https://www.w3schools.com/howto/howto_js_sidenav.asp, this is a tutorial, not a multi-purpose library, so I would not use it as a reference.

Based on the above evidence, I propose to use to `.uni-layout-left-nav` or `.uni-layout-left-side-nav`.

@PatrykSzwer can you make the suggested improvements here pls?

Issue #51, opened 2018-01-15
🤸 Add CSS variable presets


warpech:
Right now, CSS variable management starts with an empty list of variables.

It would be great if we could prepopulate the list of variables by either:

1. detecting the currently registered CSS variables
2. listing known CSS variables that come from Underwear.css

Depends on https://github.com/Starcounter/Blending/pull/48 and https://github.com/Starcounter/Blending/issues/50

PatrykSzwer:
Closing as done.

PR #109, opened 2018-03-06
Underwear and uniform variables presets

PatrykSzwer:
For https://github.com/Starcounter/Blending/issues/51


PatrykSzwer:
I've removed unused anymore variables in LayoutPage.json.cs but I don't think it needs any additional reviews, so I'll just merge.

Issue #99, opened 2018-02-28
Layout group select is empty when there is only one group


PatrykSzwer:
![image](https://user-images.githubusercontent.com/25042256/36785988-12db3d44-1c85-11e8-868c-f66892787063.png)

kegor:
As I can see this might be the common problem for all `juicy-select` components:
![image](https://user-images.githubusercontent.com/17431807/36886989-0a2cc72c-1e00-11e8-8311-b4b35b76bd07.png)
@PatrykSzwer am I right that we should see in these inputs values of the `caption-text` attributes, for example `Ungrouped` for the first one and `None` for the second?

PatrykSzwer:
According to https://github.com/Juicy/juicy-select#properties section:
`Text of an empty option, empty option will not be created if the value is not set`

alshakero:
I checked and `options` array is actually empty. Why do you expect `juicy-select` to list anything?

kegor:
Guys, I also got an idea, I think we can remove the check for `this.options.length` in the following line: https://github.com/Juicy/juicy-select/blob/84274eebb24361396c2da7d3c89989be1af9488f/juicy-select.html#L50 because in the opposite case we will never see the empty option with `caption text`.

@PatrykSzwer, @alshakero wdyt?

alshakero:
@kegor that's neat. I tried it locally and it works! I'll PR to `juicy-select` soon.

alshakero:
Created https://github.com/Juicy/juicy-select/pull/20

kegor:
@PatrykSzwer I think that current bug could be closed after PR from Omar will be merged. Wdyt?

warpech:
This bug can be closed when a new version of `juicy-select` is released and installed in BlendingEditor.

alshakero:
Heads up: [`juicy-select` is released]( https://github.com/Juicy/juicy-select/releases/tag/1.1.3).

kegor:
Closed because the fix for `juicy-select` was merged into BlendingEditor in #110.

PR #110, opened 2018-03-06
Problem #99: juicy-select should be updated in BlendingEditor.

kegor:
**Problem #99:**
The root cause of #99 was fixed in https://github.com/Juicy/juicy-select/pull/20.
So last step to fix this in BlendingEditor is to update juicy-select on version (1.1.3) which contains this fix.

**Solution:**
Juicy-select (and some other dependencies) was updated.

**Tests:**
Manually tested on running application.

**Screenshots:**
image


alshakero:
I checked the updated deps and all seems fine.

Issue #107, opened 2018-03-06
Use new classes from uniform in BlendingEditor


PatrykSzwer:
In latest release we added new classes like `uni-layout-left-nav` that can be and should be used in BlendingEditor.

PatrykSzwer:
Closing as done.

PR #112, opened 2018-03-07
Use uni classes

PatrykSzwer:
For #107

See commit message for details.


src/BlendingEditor/wwwroot/BlendingEditor/views/LayoutConditionsPage.html
```diff
@@ -188,11 +187,10 @@ <h5>Choose layout for layout condition</h5>
     </script>
     <template is="declarative-shadow-dom">
         <link rel="stylesheet" href="/sys/uniform.css/uniform.css">
-        <link rel="stylesheet" href="/BlendingEditor/css/blendinguniform.css">
         <div class="uni-section-secondary uni-title">
             <slot name="blendingeditor/layout-conditions-header"></slot>
         </div>
-        <div class="uni-section-primary blendingeditor-section-primary-grid">
```
warpech: Removal of this class results in all elements stacking without gap between them.

Expected: there should be a gap between `uni-alert-info` and `uni-card`, also between `uni-card` and the bottom buttons.

Actual:

image

Issue #121, opened 2018-03-12
Reduce the number of default layouts, add area items and layout conditions to predefined layouts from a template


warpech:
### Expected result

As a user of BlendingEditor, I would like to be able to use predefined layouts from template to recreate the layouts that BlendingEditor creates in an empty database.

The expected predefined layouts from a template are: `Single point` layout, `Topbar` layout, `Sidebar` layout (merge with current `Responsive` layout), `HolyGrail` layout, `Login` layout (data includes Layouts, Areas, Area items, Layout conditions)

The expected default layouts are: `Topbar` layout, `Login` layout (data includes Layouts, Areas, Area items, Layout conditions)

### Actual result

The current predefined layouts from a template are: `Single point` layout, `Topbar` layout, `Sidebar` layout, `HolyGrail` layout (data includes only Layouts and Areas; NOT Area items and Layout conditions)

The current default layouts are: `Topbar` layout, `Login` layout, `Sidebar` layout, `Responsive` layout, `HolyGrail` layout (data includes Layouts, Areas, Area items, Layout conditions)

I have a preliminary fix for that, but it will need more work.

warpech:
Fix is tracked in PR #122

PatrykSzwer:
Closing as done.

PR #122, opened 2018-03-12
Fix for #121: Reduce the number of default layouts, add area items and layout conditions to predefined layouts from a template

warpech:
Reuse DataHelper when creating a predefined layout from a template (#121) 

Reduce the number of default layouts created when the app starts on an empty db.

This PR is a preliminary work for #121, it needs to be finished:

- add a data helper for Single point layout and Login layout; use that data helper in the modal window of choosing a predefined layout from a template
- create an icon for the modal window for Login layout
- merge Responsive layout into Sidebar layout and remove Responsive layout from code


PatrykSzwer:
I finished all listed points in first post. It's ready to be reviewed @kegor.

kegor:
The app is runnable locally and I see new SVG-icon in Login Layout, but for the remaining points (p1 and p3) I find it difficult to verify.

PatrykSzwer:
Those are mostly about code changes so you should just check if code looks fine and all touched functionalities are still working.

PR #123, opened 2018-03-12
Change the method of detecting Uniform.css classes

warpech:
Change the method of detecting Uniform.css classes from parsing `:root` to a RegExp, because the detection of Uniform.css classes only found`--uni-default-*` custom properties.

**Caution**: this fails with shorthand rules in a CSS file that have two values in the same line, such as:

```css
padding: var(--uni-section-padding-vertical, var(--uni-default-section-padding-vertical)) var(--uni-section-padding-horizontal, var(--uni-default-section-padding-horizontal));
```

@PatrykSzwer could you please improve the regexp or change the solution to use a **client side** CSS parser, possibly by iterating over `document.styleSheets` in JavaScript, which has that information:

screen shot 2018-03-12 at 08 31 32

Maybe you can find some useful parsing libraries on NPM, such as: 

- https://www.npmjs.com/package/css-customproperties-parser (this looks naive)
- https://www.npmjs.com/package/postcss-extract-custom-properties (this can be excellent, but I am afraid that PostCSS itself may be huge)


PatrykSzwer:
I've added code for detecting those problematic variables, which should work with current version of uniform, but in the future I think it should be improved. We can create an issue about it and come back to it when it will be needed. WDYT?

warpech:
LGTM, merging.

I will not create an issue about improving it until we know it must be improved.

PR #125, opened 2018-03-12
Make the table in CSS collection page full-width

warpech:
Make the table in CSS collection page full-width because it did not fit long custom property names or values


PatrykSzwer:
image
Reference to what I meant by this position.

warpech:
Thanks @PatrykSzwer, now it looks like:

image

I have pushed a fix and will merge it if the build passes.

warpech:
Test passed, merging

PR #126, opened 2018-03-12
Use negative margins on the composition editor panel

warpech:
Use negative margins on the composition editor panel because the panel was offset by Underwear.css `8px` margin, causing the drag handle in the bottom right corner to be inaccessible


src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-panel/compositioneditor-panel.html
```diff
@@ -21,6 +21,7 @@
                 background: #FFF;
                 font-size: initial;
                 color: initial;
+                margin: -8px -8px 8px -8px;
```
kegor: minor: this can be collapsed to `-8px -8px 8px`.

kegor:
@warpech, maybe this change should be made in github.com/Starcounter/starcounter-layout-html-editor?

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-panel/compositioneditor-panel.html
warpech: I am not a big fan of a triple shorthand notation in CSS. I never know what does it mean :)

warpech:
The variant "with new margins" looks exactly how it looked before https://github.com/Starcounter/underwear.css/issues/24. 

Maybe let's merge it as is. In future, we might consider a better design.

> @warpech, maybe this change should be made in github.com/Starcounter/starcounter-layout-html-editor?

It is better if the container (`compositioneditor-panel`) takes care of positioning its children, not the positioned element (`starcounter-layout-html-editor`) itself.

warpech:
Thanks for the review!

PR #128, opened 2018-03-12
17 drag and drop routes

alshakero:
I used ordering buttons, they're easy to reason about, easy to implement and don't require any libs. Albeit I agree DnD can be more satisfying.

Fixes: https://github.com/Starcounter/Blending/issues/17


warpech:
As we discussed on the today's meeting, there is value in drag and drop. @alshakero will try again looking how https://codetree.com/ implemented drag and drop of a `<tr>`.

alshakero:
Readt for review. It works on Chrome, FF and Edge.

PatrykSzwer:
I would also experiment a little with different icon for drag. Something more like this maybe: 
image


or I think most common ways, couple of grey dots, like here:
image

alshakero:
Can you continue on your suggestions now that the DnD annoying part is done?

PatrykSzwer:
I fixed the logical part of the functionality. There is still some problem with "jumping" table, @alshakero could you take a look?
rep

alshakero:
@PatrykSzwer should be done.

PatrykSzwer:
There were some problems caused by newest changes in SC which I fixed in last commit, but turns out that there is still some problem with proper order setting with drag and drop:
rep

PatrykSzwer:
All the backend part of this feature should be working fine. Only thing left is to get keys of drag and drop items properly, after table was changed in same transaction. @alshakero will continue work on this issue.

warpech:
On today's meeting @alshakero suggested that the algorithm needs some more work. @kegor will take the C# part.

tomalec:
It seems we are almost there, but it's hard to debug it given https://github.com/Starcounter/Home/issues/121#issuecomment-376512275

Let's wait with further work, until we can actually inspect anything in VS.

PatrykSzwer:
I've fixed drag and drop feature for all layout conditions view. Please check if it works without any issues.

I talked with @warpech and we decided to, for now, drop ordering feature for specific layout view. This issue took already too much time and contains wide, conflictable changes that should be merged. 

@alshakero please take a final look and merge if you decide that It's working without any issues.

alshakero:
Works like a charm!

alshakero:
Any idea why tests are failing?

kegor:
@alshakero seems like in latest run tests was successful again, but I'm looking at this situation today.

Issue #118, opened 2018-03-09
Implement partial matching of HTTP headers


warpech:
### Expected result

As a user of BlendingEditor, I would like to define an HTTP header rule for a layout condition using partial string matching of an HTTP header. 

Let's discuss what would be better? Partial string matching rules like (begins with, contains, ends with) or regexp? I think regexp might be too slow for such usage.

### Actual result

Right now the HTTP header value is matched exactly against the stored value (case insensitive).

### Screenshot

![image](https://user-images.githubusercontent.com/566463/37205398-eefb492a-2394-11e8-80e5-1dbcabb76078.png)

PatrykSzwer:
Closing as done.

Issue #106, opened 2018-03-05
Some tests are failing on master branch build


PatrykSzwer:
https://teamcity.starcounter.org/viewLog.html?buildId=48883&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24

There is a test that seems to be flaky or just failing. Please investigate.

kegor:
Closed because fixes were merged in #119.

PatrykSzwer:
Reopening since there are still some flaky tests: https://teamcity.starcounter.org/viewLog.html?buildId=49813&tab=buildResultsDiv&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24

PatrykSzwer:
Seems like updating selenium libraries in TC fixed flakiness. I'll close this issue for now, since there is no discovered flaky tests anymore.

kegor:
I see errors on `master` branch in BE tests: https://teamcity.starcounter.org/viewLog.html?buildId=50862.
I think changes from this commit could fix the problem: https://github.com/Starcounter/Blending/commit/16c9f3ad73ed0abee10ecc561beb90c93b2551c0.

@warpech what do you think?

kegor:
Closed again, because merged changes in #151 makes sense.

PR #129, opened 2018-03-14
Partial matching for layout headers

PatrykSzwer:
For #118

**Needs https://github.com/StarcounterApps/Simplified/pull/110 to work.**


src/BlendingProvider/Website/Api/ContentHandlers.cs
```diff
@@ -355,7 +355,31 @@ private bool FindUrlByHeaders(WebUrl webUrl, Dictionary<string, string> requestH
             {
                 return true;
             }
-            return webUrl.Headers.All(wh => requestHeaders.Any(rh => rh.Key.Equals(wh.Name, StringComparison.InvariantCultureIgnoreCase) && rh.Value == wh.Value));
+
+            if (!webUrl.Headers.Any(x => x.PartialMatching))
```
un-tone: This statement looks redundant. The same is doing below.

PatrykSzwer: TBH I added this line there, to save some performance, but now when I think of it, probably version with partial matching will be even more (or at least same amount of time) used as an old one.
I'll remove this one.

kegor:
BlendingEditor is runnable locally with the change in Simplified, but at night I only can't quickly imagine the test-case for checking the exact functionality, @PatrykSzwer could you please give me a hint?

PatrykSzwer:
You can just add header rule to topbarLayout, lets say host: localhost and check how its working with/without partial matching.

But TBH I shouldn't be the one to suggest you how you should test it, because you could find new bug that I may not find trough my testing process.

PatrykSzwer:
I updated package with simplified. After tests pass I'll merge.

PatrykSzwer:
There are some tests that failed but I talked to @kegor and those are not related to my changes (new flaky tests). I'll merge this and #106 should be reopened.

PR #130, opened 2018-03-15
Adaption for upcoming breaking changes in TypedJSON

chrhol:
Added codebehind property that is bound to the property in JsonByExample to get the state of the transaction instead of having the transaction as dataobject.

The reason for this is the decoupling of TypedJSON from Starcounter 2x, and as a consequence of that the starcounter2x specific transaction `Starcounter.Transaction` cannot be directly used by casting `json.Transaction` (which have also been renamed btw, but this is a later non-breaking thing).

These changes will work on the current Starcounter version so they can be merged before the actual decoupling is merged.

I haven't actually succeeded in doing a passing build in teamcity, so that part still needs to be done.

Not sure who should be added as reviewer, so feel free to add or remove anyone.


chrhol:
Succesful build: https://teamcity.starcounter.org/viewLog.html?buildId=50011&tab=buildResultsDiv&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24

ping @warpech, @kegor, @PatrykSzwer

chrhol:
> @chrhol do you have a link to some GH issue to track the braking change in Typed JSON?

https://github.com/Starcounter/level1/issues/4007. 

I will write an announcement today as well, and push the changes to level1 later tonight.

Issue #116, opened 2018-03-09
Area item URI combo box should contain URIs from all apps


warpech:
### Expected result

Area item URI combo box should be prepopulated with available view URIs from all currently running apps.

### Actual result

It is prepopulated only from view URIs from the current app (BlendingEditor)

![image](https://user-images.githubusercontent.com/566463/37193364-c7b2f776-2369-11e8-830c-f18f81ea1a8b.png)

PatrykSzwer:
Closing as done.

PR #131, opened 2018-03-15
All uris from all registered apps

PatrykSzwer:
For #116


PatrykSzwer:
Those are created during generation process of LoginLayout and are expected (those are even used by default in login layout)

Issue #12, opened 2018-01-08
Smooth permanent deleting surfaces


PatrykSzwer:
_Issue moved from Website repository: https://github.com/Starcounter/Website/issues/267_

Based on the comment https://github.com/Starcounter/Website/issues/200#issuecomment-341848568:

1. user clicks the "Delete" button, which deletes the surface
2. this triggers a delete commit hook in WebsiteProvider
2. WebsiteProvider unmaps the surfaces before deleting
3. WebsiteProvider repeats the function https://github.com/Starcounter/Website/blob/a69aa8b94c11bb94806416b224a2b40974e87343/src/WebsiteProvider/Api/ContentHandlers.cs#L79 for all current sessions. This should remove the surface from being used because it is unmapped now
4. now the surface can be safely deleted

kegor:
The Idea in first approximation:
1. Wrap the `response` here https://github.com/Starcounter/Website/blob/a69aa8b94c11bb94806416b224a2b40974e87343/src/WebsiteProvider/Api/ContentHandlers.cs#L131 in some kind of `MasterPageLayout`.
2. Extract method from this code https://github.com/Starcounter/Website/blob/a69aa8b94c11bb94806416b224a2b40974e87343/src/WebsiteProvider/Api/ContentHandlers.cs#L79.
3. During Layout deletion find in the Session original response from which we need to find `requestUri`.
4. Using 'original' `requestUri` from p.3. we need to invoke the method from p.2.
5. Finally, we need to ScheduleTask for the session with `session.CalculatePatchAndPushOnWebSocket();`- method invocation within.

kegor:
Current state:
![layouts_smooth_deleting_1](https://user-images.githubusercontent.com/17431807/37436804-dd493fe2-27fa-11e8-943d-046b77730523.gif)

kegor:
Currently, this blocked by https://github.com/Starcounter/level1/issues/4489.

PR #134, opened 2018-03-19
Problem #12: layouts smooth deleting.

kegor:
**Problem #12:**
We can refresh views in all tabs after deleting layout in all tabs.

**Solution:**
1. Was added new `MasterWrapper` view for storing changing views on a page.
2. Was extracted method `WrapResponse`.
3. Were added handlers for BE and BP which mapped on the same token.

**Tests:**
1. Existing tests were updated.
2. 🕙 Working on new tests in multiple tabs.

**Screenshots:**
layouts_smooth_deleting_2


src/BlendingEditor/Website/Api/MainHandlers.cs
```diff
@@ -124,6 +124,8 @@ public void Register()
                 var json = new Editor();
                 return json;
             });
+
+            Handle.GET("/blendingeditor/layouts/postdelete", () => new Json());
```
un-tone: I guess this can be defined with `SelfOnly = true` option.

src/BlendingProvider/Website/Api/ContentHandlers.cs
```diff
@@ -36,6 +39,22 @@ public void Register()
...
+
+                    if (masterWrapper != null)
+                    {
+                        this.WrapResponse(masterWrapper.Request, masterWrapper.Response);
+                        session.CalculatePatchAndPushOnWebSocket();
+                    }
+                });
+
+                return null;
+            });
```
un-tone: This handler can be defined with `SelfOnly = true` option.

src/BlendingProvider/Website/Api/ContentHandlers.cs
```diff
@@ -93,12 +112,81 @@ public void Register()
...
+                {
+                    this.ResponseStorage.Remove(responseKey);
+                }
+
+                wrapper = response.Resource as LayoutPage;
+                requestUri = wrapper?.Data.Html;
+
+                if (string.IsNullOrEmpty(requestUri))
+                {
+                    return lastResponse;
```
un-tone: This response is not wrapped by `MasterWrapper`, but should be, right?
I'd recommend to extract the wrapping into a separate method and use it here and below.

un-tone:
Also, you should add new tests covering the new feature.

warpech:
Still some tests are failing: 

```
MasterWrapperTests(Chrome,DisableViewKeeper).DeleteTopbarLayout_UsingPageButton
MasterWrapperTests(Chrome,EnableViewKeeper).DeleteTopbarLayout_UsingPageButton
```

@kegor can you take a look?

kegor:
@warpech, yes, sure, but I think it makes sense after finishing the investigation in this bug https://github.com/Starcounter/Blending/issues/12#issuecomment-375917153.

kegor:
@un-tone wdyt?

src/BlendingEditor/blend.json
```diff
@@ -5,5 +5,12 @@
...
      "AllowFromDirection": false,
      "AllowToDirection": true
-  }
+  },
+  {
+    "Uri": "/blendingeditor/layouts/postdelete",
+    "Token": "layout-post-delete",
+    "Contexts": [],
+    "AllowFromDirection": true,
+    "AllowToDirection": true
```
un-tone: I guess `AllowToDirection` can be changed to `false` since `/blendingeditor/layouts/postdelete` handler does nothing.

src/BlendingProvider/Website/Api/BlendingMaps.cs
```diff
@@ -0,0 +1,19 @@
...
+        public BlendingMaps()
+        {
+        }
+
+        /// <summary>
+        /// Registers the mapping for Uri and token
+        /// </summary>
+        public void Register()
+        {
+            Blender.MapUri("/BlendingProvider/viewmodels/refresh", "layout-post-delete");
```
un-tone: You can optimize it with direction parameters
```cs
Blender.MapUri("/BlendingProvider/viewmodels/refresh", "layout-post-delete", false, true);
```

Issue #141, opened 2018-03-22
"Delete all" button in settings should delete layout groups as well


mmnosek:
In `BlendingEditor`'s settings page, there are two red "Delete all" buttons: in layouts and compositions sections. I think that it will be useful if "Delete all layouts" removes layout groups as well.

kegor:
Closed, because the fix was merged in the #152.

PR #152, opened 2018-03-27
Problem #141: "Delete all" button in settings should delete layout groups as well

kegor:
**Problem #141:**
> In BlendingEditor's settings page, there are two red "Delete all" buttons: in layouts and compositions sections. I think that it will be useful if "Delete all layouts" removes layout groups as well.

**Solution:**
Deletion of the layout groups was added in the `DeleteLayoutsModalTrigger` handler method.

**Tests:**
✅ The manual check was performed.

**Screenshots:**
delete_all_layouts_and_groups


src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -37,14 +37,22 @@ void Handle(Input.CancelModalTrigger action)
...
+
             Db.Transact(() =>
             {
                 var layouts = Db.SQL<WebTemplate>($"SELECT wt FROM {typeof(WebTemplate).FullName} wt");
                 foreach (WebTemplate layout in layouts)
                 {
                     layout.Deleted = true;
                 }
+
+                var layoutsGroups = Db.SQL<WebTemplateGroup>($"SELECT wtg FROM {typeof(WebTemplateGroup).FullName} wtg");
```
un-tone: Using the `FullName` property is redundant here. Just use `{typeof(WebTemplateGroup)}`

un-tone: I see now that above the same approach is used. Please change there too while we are here.

PR #154, opened 2018-03-28
6.1.0

kegor:
**Problem:**
We need to make a release for Blending.

**Solution:**
Versions of BlendingProvider and BlendingEditor assemblies were changed from 6.0.0 to 6.1.0.


kegor:
@alshakero could you please take a look at the release notes which I added here https://github.com/Starcounter/Blending/releases/edit/untagged-755b867637043fcaebcb. Are they ok for you?

Issue #137, opened 2018-03-20
We need to warn a user if he trying to save Layout Condition with empty URI.


kegor:
### Steps to reproduce
1. Create new Empty Layout.
2. Go to the 'Layout conditions'.
3. Click 'New condition'.
4. Click 'Save' (without changing any inputs).
5. Go to the 'General'.

### Expected result
A warning message about empty URI that leads to the 'default' applying of this Layout for all URIs.

### Actual result
Empty white view with only TopBarLayout.

##### Starcounter version 
2.4.0.5313

##### Involved apps' version
- BlendingEditor@6.0.0
- BlendingProvider@6.0.0
 
##### Browsers affected
- [x] Chrome
- [x] Firefox

### Screenshot
![image](https://user-images.githubusercontent.com/17431807/37649624-66d51136-2c43-11e8-96d3-e44762765853.png)

warpech:
I was wrong with the idea to do any kind of "validation" for empty layout condition URI. It cannot be validated, because an empty value is also valid.

I think that the layout needs to be changed.

1. "URI" should be a button, similarly to how "Headers" is a button
2. The button should display the URI as the value, or "Any URI" if the value is an empty string
3. Remove the title from `th` that represents the URI in the table
4. Move the input to set the URI into the expanded row
5. Move "Use regular expressions" into the expanded row
6. Add info about wildcards and empty URIs as `uni-alert-info` in the expanded row

A visualization of what I mean (don't take it as 1:1 design, just as a guideline):

![all routes uri expanded](https://user-images.githubusercontent.com/566463/38197996-3219d47a-368b-11e8-8ba9-8c408cf3b09c.png)

@kegor WDYT?

kegor:
- I got the idea that empty value is also valid, but there is still the case in which user can get the white screen after addition such layout condition. This is not so good in my view.
- regarding the new UX guideline, it looks clean, I'd maybe only change the info-block regarding `any URI` to yellow.

kegor:
@warpech,
1. Default area: I've updated test case and screenshot in task description, even with default area I'm getting white screen, but we should get usual page here, and this looks like yet another bug?
2. Sounds good.

warpech:
> Default area: I've updated test case and screenshot in task description, even with default area I'm getting white screen, but we should get usual page here, and this looks like yet another bug?

Sounds like a bug.

PR #155, opened 2018-03-28
Problem #137: warning for a user if he trying to save Layout Condition with empty URI.

kegor:
**Problem #137:**
 Will be good to warn a user if he trying to save Layout Condition with empty URI.

**Solution:**
The warning that is triggered on back-end was added.

**Tests:**
✅ The manual check was successful.

**Screenshots:**
warning_for_layoutcondition_with_empty_uri


un-tone:
I guess, it is not enough to just warn, preventing is better for this case. Otherwise, the layout condition is saved and if a user will go to any other page, i.e. "General" he still gets an empty page and he will be able to do nothing with it. It will be a big problem for the user if he had some working settings there.

un-tone:
I should note that a user is getting an empty page only in case if a user added a condition for newly added layout. I propose to implement additional checkings to prevent adding such breaking data.

warpech:
I agree with @un-tone that this needs changes. I will propose a different UX design for it.

warpech:
I added my proposal of changes to https://github.com/Starcounter/Blending/issues/137#issuecomment-377930123.

If you agree then I think this PR should be closed and a new one is needed.

Big thanks for being proactive in proposing such solutions, anyway!

kegor:
Closed because we decided to improve UX of the Layout Conditions page https://github.com/Starcounter/Blending/issues/137#issuecomment-377930123. And the implementation of the warning will change.

Issue #142, opened 2018-03-22
Add "Reset to defaults" button in settings


mmnosek:
When I install `BlendingEditor` for the first time, I have some default layouts that look very nice. After playing around, I would like to be able to reset everything to the initial (post-installation) state. Currently, if I delete all layouts and compositions, I am stuck. 

Red "Reset to defaults" button should:
* delete layouts (same operation as red delete all button)
* delete compositions
* reset "use view keeper" setting
* create the same objects as on app installation (or first run, actually)
* be secured by modal, with appropriate warning message

kegor:
Closed because the fix was merged in #158.

Issue #140, opened 2018-03-22
"Use view keeper" setting isn't exported to configuration dump


mmnosek:
### Steps to reproduce
1. Uncheck "Use view keeper" in the configuration (by default it's checked)
1. Export data
1. Check "Use view keeper"
1. Import data
1. Refresh

### Expected result
"Use view keeper" checkbox shouldn't be checked

### Actual result
"Use view keeper" checkbox is checked, meaning it was not imported or exported

kegor:
For me Import even is not starting:
![image](https://user-images.githubusercontent.com/17431807/38131825-26ef7c7a-3411-11e8-9220-b151350aacfd.png)
I will investigate this.

kegor:
I found the solution for the problem above (changing `as` on `is`), but the whole fix has the problem related to this case:
https://github.com/Starcounter/Blending/issues/140
![image](https://user-images.githubusercontent.com/17431807/38237513-135e179e-3730-11e8-8433-0d16893e8b22.png)

kegor:
Closing because related changes were merged in the #253.

PR #158, opened 2018-03-29
Problem #142: Add "Reset to defaults" button in settings.

kegor:
**Problem #142:**
> When I install BlendingEditor for the first time, I have some default layouts that look very nice. After playing around, I would like to be able to reset everything to the initial (post-installation) state.

**Solution:**
Such button was added. It now uses existing methods from DataHelper which also could be invoked using URL: `/resetdata`.

**Test:**
Manual test that is presented in screenshots was successful.

**Screenshots:**
reset_to_defaults_button


src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -25,7 +25,13 @@ void Handle(Input.OpenLayoutsModalTrigger action)
 
         void Handle(Input.OpenCompositionsModalTrigger action)
         {
-            this.IsLayoutsModalOpened = false;
+            this.IsCompositionsModalOpened = true;
```
warpech: When setting `IsCompositionsModalOpened` to `true`, you need to set other flags to `false`.

Otherwise, if you first delete all compositions, then want to reset to defaults, you get both messages and both buttons.

Perhaps, a better idea would be to have just two properties in JSON: `ModalMessageText` and `ModalButtonText`.

2018-04-02 12_08_19-blendingeditor

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -63,6 +69,14 @@ void Handle(Input.DeleteCompositionsModalTrigger action)
             this.Message = "Compositions were deleted successfuly!";
         }
 
+        void Handle(Input.ResetToDefaultsModalTrigger action)
+        {
+            DataHelper.ClearData();
+            DataHelper.GenerateData();
+            this.IsModalOpened = false;
+            this.RedirectUrl = "/blendingeditor/settings";
```
warpech: `RedirectUrl` looks like some forced reload of the current view. But in this case, I assume it would also be needed when deleting all layouts and compositions, not? If it is only needed in this line, could you pls add some comment about what exactly needs reloading?

src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -48,24 +48,33 @@ <h3 slot="blendingeditor/settings-layouts-header">Layouts</h3>
...
             <div>
                 <template is="dom-if" if="{{model.IsLayoutsModalOpened}}">
                     <h5>Are You sure You want to delete all layouts, areas, area items and layout conditions?</h5>
                 </template>
-                <template is="dom-if" if="{{!model.IsLayoutsModalOpened}}">
+                <template is="dom-if" if="{{model.IsCompositionsModalOpened}}">
                     <h5>Are You sure You want to delete all compositions?</h5>
                 </template>
+                <template is="dom-if" if="{{model.IsResettodefaultsModalOpened}}">
+                    <h5>Are You sure You want reset to defaults?</h5>
```
warpech: In all these modal questions, `you` should be written in all lowercase according to the English language orthography.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
kegor: Yes, I agree, I also see `RedirectUrl` in [DeleteLayoutsModalTrigger](https://github.com/Starcounter/Blending/blob/82361e5bed04d004cceee293b2908cfd03a5c07a/src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs#L62), and I guess that this is really some kind of force refresh to avoid client-side errors related at least to the `ViewKeeper` check-box (I mentioned this error case [here](https://github.com/Starcounter/Blending/issues/140#issuecomment-378166702).

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -388,6 +389,15 @@ public static WebTemplate GenerateLoginLayout()
             return layout;
         }
 
+        public static void EnableViewKeeper()
+        {
+            Db.Transact(() =>
+            {
+                var settings = DataHelper.GetSettings();
+                settings.EnableViewKeeper = true;
```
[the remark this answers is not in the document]
kegor: I think I was partially wrong in the previous comment, and for now, I could propose the following approach:
1. Rename `GetSettings` method to `GetOrCreateSettings`/`EnsureWebSettings` (because of its current implementation). As far as I understand this method creates exactly default settings.
2. Add command for clearing `WebSettings` table in [`ClearData` method](https://github.com/Starcounter/Blending/blob/82361e5bed04d004cceee293b2908cfd03a5c07a/src/BlendingEditor/Website/Helpers/DataHelper.cs#L25).
2. Remove method 'EnableViewKeeper' created by me, and replace its invocation with `EnsureWebSettings` which is will creates defaults settings again.

@warpech what do you think?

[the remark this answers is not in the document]
warpech: Sounds good! What I care mostly is that resetting this setting should be DRY.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
warpech: Thanks for the investigation. Please add a comment in code about it.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
kegor: @warpech, I was thinking about your suggestion with `ModalMessageText` and `ModalButtonText` but I have not yet come up with the idea how use these properties in a buttons block, because I should change not only the button text but also button handler.

So for now, I just added few lines for proper chaning values for all existing ViewModel properties like `IsCompositionsModalOpened`.

src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -48,13 +48,19 @@ <h3 slot="blendingeditor/settings-layouts-header">Layouts</h3>
...
-                    <h5>Are You sure You want to delete all layouts, areas, area items and layout conditions?</h5>
+                    <h5>Are you sure you want to delete all layouts, areas, area items and layout conditions?</h5>
+                </template>
+                <template is="dom-if" if="{{model.IsCompositionsModalOpened}}">
+                    <h5>Are you sure you want to delete all compositions?</h5>
                 </template>
-                <template is="dom-if" if="{{!model.IsLayoutsModalOpened}}">
-                    <h5>Are You sure You want to delete all compositions?</h5>
+                <template is="dom-if" if="{{model.IsResettodefaultsModalOpened}}">
+                    <h5>Are you sure you want reset to defaults?</h5>
```
PatrykSzwer: We definitely need some additional explanation here for the user, what the actual "reset to default" means, what will be deleted/replaced etc.

src/BlendingEditor/Website/ViewModels/SettingsPage.json
```diff
@@ -15,10 +15,14 @@
 
   "OpenLayoutsModalTrigger$": 0,
   "OpenCompositionsModalTrigger$": 0,
+  "OpenResettodefaultsModalTrigger$": 0,
   "IsLayoutsModalOpened": false,
+  "IsCompositionsModalOpened": false,
+  "IsResettodefaultsModalOpened": false,
```
PatrykSzwer: Wrong casing, should be `IsResetToDefaultsModalOpened`

src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -48,13 +48,19 @@ <h3 slot="blendingeditor/settings-layouts-header">Layouts</h3>
         <h3 slot="blendingeditor/settings-compositions-header">Compositions</h3>
         <button slot="blendingeditor/settings-compositions-button" onmousedown="++this.value;" value="{{model.OpenCompositionsModalTrigger$::click}}">Delete all</button>
 
+        <h3 slot="blendingeditor/settings-resettodefaults-header">Reset data</h3>
+        <button slot="blendingeditor/settings-resettodefaults-button" onmousedown="++this.value;" value="{{model.OpenResettodefaultsModalTrigger$::click}}">Reset to defaults</button>
```
PatrykSzwer: We definitely need some additional explanation here for the user, what the actual "reset to default" means, what will be deleted/replaced etc.

Since we're here you could also add some short sentence to `Layouts` and `Compositions` deleting cards in settings, explaining what those will do (it will look better when every card will have same structure - title, description and button)

BTW I think that reset to default instead of defaults sounds better, but that's just my opinion. Depends how you will explain it, either as reseting all tables in database to default states, or reseting application state to It's initial one.

src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -43,18 +43,27 @@ <h3 slot="blendingeditor/settings-keeper-header">View Keeper</h3>
         <label slot="blendingeditor/settings-keeper-use-note">Use view keeper</label>
 
         <h3 slot="blendingeditor/settings-layouts-header">Layouts</h3>
+        <label slot="blendingeditor/settings-layouts-note">This action will remove all layouts and layouts groups from the database.</label>
```
PatrykSzwer: This doesn't match with the description in the modal, when you click the button:
>Are you sure you want to delete all layouts, areas, area items and layout conditions?

kegor: @PatrykSzwer I decided to shorten the warning message in the modal window because in this method I only see removing for Layouts and Layouts Groups:
https://github.com/Starcounter/Blending/blob/82361e5bed04d004cceee293b2908cfd03a5c07a/src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs#L43
wdyt?

PatrykSzwer: yes, layout will have `Deleted` property set on true, but that means it will be deleted (ur smooth deleting PR should do that) along with all its dependencies.

Issue #164, opened 2018-04-08
Error after turning off/on ViewKeeper and going to another page.


kegor:
### Steps to reproduce
1. Run BlendingProvider/BlendingEditor and SignIn.
2. Go to the `Settings` page.
3. Turn on/off ViewKeeper.
4. Go to another page.

### Expected result
There are no exceptions.

### Actual result
There is an exception.

##### Starcounter version
2.4.0.5753

##### Involved apps' version
 - Starcounter/BlendingProvider@6.1.0
 
##### Browsers affected
- [x] Chrome
- [x] Firefox

### Screenshot
![bug_after_disabling_viewkeeper](https://user-images.githubusercontent.com/17431807/38466185-f66910c4-3b2d-11e8-9510-300ff7303adc.gif)

PR #165, opened 2018-04-08
Problem: error after turning off/on ViewKeeper and going to another page.

kegor:
**Problem #164:**
Same as in the subject.

**Solution:**
Additional invocation of the DetachJsonFromParent method was added.

**Screenshots:**
switching_viewkeeper_without_errors


kegor:
@un-tone, it will be good if I'll add integration test which is repeats steps from the gif above?

un-tone:
@kegor yes

un-tone:
The solution is correct. Only test coverage is required.

Issue #18, opened 2018-01-08
Add button to close the editor


PatrykSzwer:
_Issue moved from CompositionEditor repository: https://github.com/Starcounter/CompositionEditor/issues/3_

Basically, all programs and pop-ups have a way to close the window by clicking on a close button in the top right or top left corner. It would thus be intuitive to add a simple button that does this to the editor.

warpech:
Today we made a decision to:

- add a "Close" button at the right the "Save" button
- clicking on "Close" should cancel pending changes and close the composition editor panel
- remove the "Cancel" button

tomalec:
Could you give a reason why revert (without closing) feature was removed?

Also, closing without reverting (as it was before available with key shortcut) - personally I find it useful for fiddling and debugging, but maybe that's not a common case.

warpech:
Revert without closing is considered an anti-pattern. 

It is sometimes implemented using a `<input type="reset">`, but most users have a problem understanding what it does and are likely to remove their work by mistake. Criticism of reset button:

- opinion by Jacob Nielsen, the god of web usability: https://www.nngroup.com/articles/reset-and-cancel-buttons/
- more arguments: https://ux.stackexchange.com/questions/93671/do-we-really-require-a-reset-button-for-form-fields

If you think about it, can you name one piece of popular software that has a feature to revert without closing?

tomalec:
I'd say we should rather compare not to form features but rather file editing features (like in IDE or visual tool). But, I get the point of that even there is rarely revert button and it may be more confusing that just close and reopen.

Speaking of closing without reverting, WGYT about "minimize" kind of button?

warpech:
> Speaking of closing without reverting, WGYT about "minimize" kind of button?

I think that a need for this will not be high when we can open the editor in a separate window: https://github.com/Starcounter/Blending/issues/146

kegor:
Closed because all needed changes were merged into related repositories.

PR #167, opened 2018-04-10
Problem #18: listener for the close button.

kegor:
**Problem #18:**
For the moment we still think that the 'Close' button is the mostly better idea than the 'Cancel' button.

**Solution:**
This is additional PR to this one https://github.com/Starcounter/starcounter-layout-html-editor/pull/28 and it contains listener for the 'close' event.

**Tests:**
✅ The manual test was ok.


src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -22,13 +22,20 @@
                 if (!panel) {
                     panel = document.createElement("compositioneditor-panel");
                     document.body.insertBefore(panel, document.body.firstChild);
+                    
+                    const editor = document.getElementsByTagName("starcounter-layout-html-editor")[0];
```
tomalec: If I would have to select `starcounter-layout-html-editor`, I'd make it `panel.getElementsByTagName("starcounter-layout-html-editor")[0]`.

But I think it should actually listen to the (bubbling) `close` event on `compositioneditor-panel` in case `-editor` and `-panel` get detached from each other by the third party. And not to rely on an internal implementation detail that it uses `-editor` but only rely on `-panel` which is the element that `-inserter` employs.

tomalec: Then `-panel` would have to do
```js
  (function(){
		customElements.define('compositioneditor-panel', class extends Polymer.Element{
            is: 'compositioneditor-panel'
			constructor(){
				this.forwardEvent.bind(this);
			}
            attached: function() {
                const editor = document.createElement('starcounter-layout-html-editor');
				this.editor = editor;
				this.appendChild(editor);
				editor.addEventListener('close', this.forwardEvent);
            }
            detached: function() {
				this.editor.removeEventListener('close', this.forwardEvent);
				this.removeChild(this.editor);
            }
			forwardEvent(ev){
				this.dispatchEvent(new ev.constructor(ev.type, ev));
			}
        });
    })();
```

kegor: @tomalec thank you for this improvement, I'd try to add these changes, and now -editor still opening without the errors, but it doesn't close by clicking the 'Close' button without any errors.

I checked that execution was going through this line https://github.com/Starcounter/starcounter-layout-html-editor/blob/3c4e9fc4a4286e97352c64f4259633da2f53f790/starcounter-layout-html-editor.html#L364, so seems like the needed event was dispatched. Could you please help me with ideas how to debug this process more deeper?

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -22,20 +22,13 @@
                 if (!panel) {
                     panel = document.createElement("compositioneditor-panel");
                     document.body.insertBefore(panel, document.body.firstChild);
-                    
-                    const editor = document.getElementsByTagName("starcounter-layout-html-editor")[0];
-                    editor.addEventListener('close', closePanel);
-
```
tomalec: You still need

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -22,20 +22,13 @@
...
                     setInsertersState(true);
                 }
             }
 
             function closePanel() {
                 var panel = document.querySelector("compositioneditor-panel");
                 if (panel) {
-                    const editor = document.getElementsByTagName("starcounter-layout-html-editor")[0];
-                    editor.removeEventListener('close', closePanel);
```
tomalec: and `panel.removeEventListener('close', closePanel);`

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-panel/compositioneditor-panel.html
```diff
@@ -30,8 +30,21 @@
     (function(){
         Polymer({
             is: "compositioneditor-panel",
+            constructor(){
```
tomalec: You need to use either class syntax of legacy `Polymer({` one, you cannot mix both

tomalec:
I have also converted `compositioneditor-panel` to vanilla JS custom element.
I have reviewed the source code and checked manually in FF and Chrome that it works. LGTM. 
feel free to merge if it looks and works fine for you.

PR #176, opened 2018-04-17
Problem #140: export viewkeeper setting.

kegor:
**Problem #140:**
Settings is not exporting by the StarDump.

**Solution**:
WebSettings table was added into StarDump command.
Also with @PatrykSzwer we decided to add the following logic:
1. If we have WebSettings in the dump then we should remove all old records and leave only imported.
2. Otherwise, we should leave existing WebSettings records.

**Tests:**
✅ Manual check is successful.

**Screenshots:**
import-viewkeeper-settings2


src/BlendingEditor/Website/Api/MainHandlers.cs
```diff
@@ -85,13 +85,13 @@ public void Register()
                     {
                         var master = SessionHelper.GetMasterPage();
 
-                        var page = master?.CurrentPage as SettingsPage;
-
-                        if (page == null)
+                        if (!(master?.CurrentPage is SettingsPage))
```
PatrykSzwer: Why did you added these changes here? Was there any case that this could throw an exception or something?

kegor: Yeah, this is really strange, but I noticed that with old code `page` here was always `null` and therefore import didn't work at all, I tried to find the logical answer but I didn't and I decided to rewrite this using `is`, in my view the old sense has preserved.

kegor:
@PatrykSzwer I've updated this PR while I have not ideas yet for more urgent tasks.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -588,5 +588,14 @@ private static string CreatePreviewAreaItemUri(string areaKey)
         {
             return $"/blendingeditor/area/{{{areaKey}}}/preview";
         }
+
+        public static void DeleteOldSettings()
+        {
+            Db.Transact(() =>
+            {
+                var maxObjectNo = Db.SQL($"SELECT MAX(s.ObjectNo) FROM {typeof(WebSettings)} s").First();
+                Db.SQL($"DELETE FROM {typeof(WebSettings)} WHERE ObjectNo < ?", maxObjectNo);
```
PatrykSzwer: Comparing ObjectNo doesn't seem like good idea for me. It's not meant to be used this way, and I'm not 100% sure if it will always behave in the expected way (new objectNo always bigger than last one). I would rather suggest to add new column to the table, `CreatedOn` e.g which will contain date of object creation.

kegor: After additional conversation we decided realize this using following way:
check if the import actually contains settings entity, if yes then remove the already existing one, if not then leave it be.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -588,5 +588,33 @@ private static string CreatePreviewAreaItemUri(string areaKey)
         {
             return $"/blendingeditor/area/{{{areaKey}}}/preview";
         }
+
+        public static void DeleteOldSettings(ulong[] existingObjectNumbers, bool settingsExistsInDump)
+        {
+            Db.Transact(() =>
+            {
+                var allSettings = Db.SQL<WebSettings>($"SELECT ws FROM {typeof(WebSettings)} ws");
```
PatrykSzwer: You should check for Deleted = false also, like in the rest of the queries.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -141,10 +143,13 @@ void Handle(Input.ExportDataTrigger action)
         public void ImportDataToDatabase(string filePath)
         {
             var currentSessionId = Session.Current.SessionId;
+            var settingsExistsInDump = this.SettingsExistsInDump(filePath);
+            
+            var existingSettings = Db.SQL<WebSettings>($"SELECT ws FROM {typeof(WebSettings)} ws");
```
PatrykSzwer: Also needs check for not deleted ones.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -141,10 +143,13 @@ void Handle(Input.ExportDataTrigger action)
...
+            var existingObjectNumbers = existingSettings.Select(x => x.GetObjectNo()).ToArray();
```
PatrykSzwer: Those two lines can be moved to DeleteOldSettings() method since they are not used anywhere else.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -141,10 +143,13 @@ void Handle(Input.ExportDataTrigger action)
         public void ImportDataToDatabase(string filePath)
         {
             var currentSessionId = Session.Current.SessionId;
+            var settingsExistsInDump = this.SettingsExistsInDump(filePath);
```
PatrykSzwer: I believe this also can be moved deeper, It's not needed here.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -588,5 +588,33 @@ private static string CreatePreviewAreaItemUri(string areaKey)
         {
             return $"/blendingeditor/area/{{{areaKey}}}/preview";
         }
+
+        public static void DeleteOldSettings(ulong[] existingObjectNumbers, bool settingsExistsInDump)
```
PatrykSzwer: I think that awesome idea would be to return boolean here, to indicate if old settings were replaced or not(if there ware no settings in the dump file), and then based on it, change the message to say something like: "Import completed! New settings are loaded!" or similar.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
kegor: In my idea here these two line taking all existing settings before any changes by dumper.
And in `DeleteOldSettings` I'm trying to get all existing settings after Dumper make his work.
But maybe I'm wrong in this logic, wdyt?

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
kegor: Yeah, in the fact we can move this deeper, but then we need pass deeper the `filePath`, this why I decided maybe we could examine the dump file right here?

src/BlendingEditor/Website/Helpers/DataHelper.cs
kegor: User friendly idea 👍:
image

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
PatrykSzwer: I would just pass `filePath` deeper and not waste time on that part anymore.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
PatrykSzwer: Yes, you're right. Ignore my comment, this is done in a correct way.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -604,6 +606,7 @@ public static void DeleteOldSettings(ulong[] existingObjectNumbers, bool setting
                         if (existingObjectNumbers.Contains(objectNo))
                         {
                             Db.SQL($"DELETE FROM {typeof(WebSettings)} WHERE OBJECTNO = ?", objectNo);
+                            result = true;
```
PatrykSzwer: `result` -> `wasDeleted` e.g, result is too ambigius.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -163,9 +163,10 @@ public void ImportDataToDatabase(string filePath)
                     if (!this.IsErrorMessage)
                     {
                         DataHelper.RefreshLayoutUris();
-                        this.Message = "Import completed!";
+                        var result = DataHelper.DeleteOldSettings(existingObjectNumbers, settingsExistsInDump);
 
-                        DataHelper.DeleteOldSettings(existingObjectNumbers, settingsExistsInDump);
+                        this.Message = "Import completed!";
+                        if (result) this.Message += " New settings are loaded!";
```
PatrykSzwer: Never use conditions one-line and without `{}`. It's bad practice.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -588,5 +589,64 @@ private static string CreatePreviewAreaItemUri(string areaKey)
         {
             return $"/blendingeditor/area/{{{areaKey}}}/preview";
         }
+
+        public static bool DeleteOldSettings(ulong[] existingObjectNumbers, string filePath)
```
PatrykSzwer: This method is placed between private methods on the bottom. The order should always be first public methods then Private ones.  (btw It can be made private anyway, so just change access and then you don't have to move it.)

After that change I would like to wait with merge till latest SC build is fixed and tests will pass correctly.

PatrykSzwer: Sorry for the false advice, I thought that this method is in the same file (SettingsPage.json.cs) as all its callings. The part with the misplacement was still true though.

kegor:
Caught flaky problems with WebStuffUpdatingTests.

_[AddNewArea_LoadSameLayoutWithNewArea](https://github.com/Starcounter/Blending/blob/72e2a06271a4a04dd01e90194e7f1f0c74ffa4ad/test/BlendingProvider.Tests/Test/WebStuffUpdatingTests.cs#L38)_:
image

_[RecreateArea_LoadSameLayoutWithRecreatedArea](https://github.com/Starcounter/Blending/blob/72e2a06271a4a04dd01e90194e7f1f0c74ffa4ad/test/BlendingProvider.Tests/Test/WebStuffUpdatingTests.cs#L113)_:
image

PatrykSzwer:
Are those tests flaky only for this PR or It's not relevant to it?

kegor:
After this line: https://github.com/Starcounter/Blending/blob/95f982be13745ce53dd48110976c228069d6b026/src/BlendingProvider/Website/Api/BlendingHandlers.cs#L349 I put the logging of almost all `mapInfo` fields and got following log data and situation in browser like in middle screenshot above:
```
mapInfo.CallerUri: /blendingprovider/blender/layout/FN/area/FQ 15:14:12.018 38
mapInfo.IsLayoutConditionAssigned: False 15:14:12.018 38
mapInfo.TokenBase: blendingprovider%FN%FQ 15:14:12.018 38
mapInfo.ForeignUri: /BlendingProvider_AcceptanceHelperTwo/pin1 15:14:12.018 38
mapInfo.SectionId: 336 15:14:12.018 38
mapInfo.Token: blendingprovider%FN%FQ 15:14:12.018 38
mapInfo.SectionId: 337 15:14:12.020 38
mapInfo.TokenBase: blendingprovider%FN%FR 15:14:12.020 38
mapInfo.CallerUri: /blendingprovider/blender/layout/FN/area/FR 15:14:12.020 38
mapInfo.Token: blendingprovider%FN%FR 15:14:12.020 38
mapInfo.IsLayoutConditionAssigned: False 15:14:12.020 38
mapInfo.ForeignUri: /BlendingProvider_AcceptanceHelperTwo/pin2 15:14:12.020 38
mapInfo.CallerUri: /blendingprovider/blender/layout/FN/area/FP 15:14:12.021 38
mapInfo.SectionId: 335 15:14:12.021 38
mapInfo.Token: blendingprovider%FN%FP 15:14:12.021 38
mapInfo.TokenBase: blendingprovider%FN%FP 15:14:12.021 38
mapInfo.ForeignUri: /BlendingProvider_AcceptanceHelperTwo/pin3 15:14:12.021 38
mapInfo.IsLayoutConditionAssigned: False 15:14:12.021 38
mapInfo.CallerUri: /blendingprovider/blender/layout/FN/area/Fq 15:14:13.996 34
mapInfo.Token: blendingprovider%FN%Fq 15:14:13.997 34
mapInfo.ForeignUri: /BlendingProvider_AcceptanceHelperTwo/pin4 15:14:13.997 34
mapInfo.TokenBase: blendingprovider%FN%Fq 15:14:13.997 34
mapInfo.SectionId: 362 15:14:13.997 34
mapInfo.IsLayoutConditionAssigned: False 15:14:13.997 34
```
— so we can see that tokens for pin1 and pin4 are very similar except of register of one symbol. So my idea is that we have this flaky errors in these tests because of case-insensitive in tokens, perhaps it was added recently.

warpech:
This should be closed because it uses the old dump file format.

@kegor can you pls make a new PR for the JSON based file format.

kegor:
Closing because new PR for the new JSON-migration should be created.

Issue #160, opened 2018-04-04
Only insert blendable URIs as area items


warpech:
From: https://github.com/Starcounter/Launchpad/issues/40#issuecomment-378754522

BlendingProvider should ~~not~~ skip inserting the area item if the URI is not listed in `Blender.ListAllByUris()`. 

BlendingEditor should warn that this URI is not found in the list of blendable URIs, like this:

![all routes uri expanded](https://user-images.githubusercontent.com/566463/38337360-a4ab85ae-386d-11e8-8221-e1f60c7a4b99.png)

PatrykSzwer:
I've done the second part, but TBH I don't really understand the first:
> BlendingProvider should *not* skip inserting the area item if the URI is not listed in Blender.ListAllByUris().

As far as I know, there is no such check in BlendingProvider. Could you clarify more what you meant by that sentence?

warpech:
First of all, sorry for the typo. I meant that it **should skip inserting the area item**.

I mean this method of code: https://github.com/Starcounter/Blending/blob/b98142170b56604f4f295f77858ff9e2d5f80001/src/BlendingProvider/Website/Api/BlendingHandlers.cs#L263

Currently, it sets a blending rule to any `webMap.ForeignUri`. My proposal is to validate if `ForeignUri` registered as blendable by checking if:

- it is present in `Blender.ListAllByUris()` (for example, People app is running and `ForeignUri` equals `/people/partial/person/{?}`)
- it is a mixed URI for a URI that is present in `Blender.ListAllByUris()` (for example, People app is running and `ForeignUri` equals `/people/partial/person/{123}`)

PatrykSzwer:
Actually, just fixing the typo would be explanatory enough 😃 But thank you for good explanation, now I know what you meant.

PatrykSzwer:
One problem came up about this issue: If we skip inserting the area items in BlendingProvider method, then even if user turns on his app with proper URI, It will still not be mapped since currently, as @alemoi told me, there is no indicator that would say, whenever blending rules have changed. 

The best solution right now will be to add a warning to UI when user wants to add URI which doesn't exist in `Blender.ListAllByUris()`, and later when we move blending rules to the database, solution for that should be fairly easily obtainable.

PatrykSzwer:
Closing as done.

PR #178, opened 2018-04-20
Warning about adding unlisted URI

PatrykSzwer:
For #160.


src/BlendingEditor/Website/ViewModels/AreasPage.json.cs
```diff
@@ -131,6 +135,11 @@ void Handle(Input.UrlKey action)
             {
                 this.Data.Url = string.IsNullOrEmpty(action.Value) ? null : Db.FromId<WebUrl>(action.Value);
             }
+
+            void Handle(Input.ForeignUrl action)
+            {
+                this.IsUriRegistered = registeredUris.Any(x => x == action.Value);
+            }
```
warpech: Wouldn't it be DRYer to do `IsUriRegistered` as a getter function?

Then:

```
void Handle(Input.ForeignUrl action)
            {
                this.IsUriRegistered = registeredUris.Any(x => x == action.Value);
            }
```

src/BlendingEditor/wwwroot/BlendingEditor/views/AreasPage.html
```diff
@@ -54,6 +54,13 @@ <h2 slot="blendingeditor/areas-layout-header">Areas - {{model.Layout.Name}}</h2>
                     <template is="dom-if" if="{{area.local.IsExpanded}}" restamp>
                         <tr>
                             <td colspan="8" style="background-color: var(--mouse);">
+                                <template  is="dom-repeat" items="{{area.AreaItems}}" as="areaItem">
+                                    <template is="dom-if" if="{{!areaItem.IsUriRegistered}}">
+                                        <div class="blendingeditor-uri-warning">
+                                            Warning: URI <code>"{{areaItem.ForeignUrl$}}</code> couldn't be found on registered URIs list. Make sure that the app is started and the URI is handled.
+                                        </div>
+                                    </template>
+                                </template>
```
warpech: I think it would make sense to put this validation message just below `<vaadin-combo-box>`, in the same TD as the combo box.

You can even use leverage built-in styling of `vaadin-combo-box` by setting the attribute `invalid` on that element. See: https://vaadin.com/components/vaadin-combo-box/html-api/elements/Vaadin.ComboBoxElement

src/BlendingEditor/Website/ViewModels/AreasPage.json.cs
PatrykSzwer: Yes you're right. I was trying to implement this as a getter at first, but then after some troubles, I gave up the idea. Now at second try, I've managed to do it.

src/BlendingEditor/wwwroot/BlendingEditor/views/AreasPage.html
PatrykSzwer: TBH I think that current solution with message above table is much better looking, I like the idea with the attribute though.

Issue #177, opened 2018-04-17
Switch from Blender.MapUri to blend.json in BlendingEditor


warpech:
Switch from `Blender.MapUri` to `blend.json` in BlendingEditor.

Docs about `blend.json`: https://docs.starcounter.io/v/2.4/topic-guides/blendable-web-apps/view-model-attaching

kegor:
Closed, because related changes were merged in #183.

PR #183, opened 2018-04-26
Problem #177: Switch from Blender.MapUri to blend.json in BlendingEditor.

kegor:
**Problem #177:**
We need to make a switch from Blender.MapUri to blend.json in BlendingEditor.

**Solution:**
Use of `Blender.MapUri` was removed.
Configuration file `blend.json` was added.

**Tests:**
✅ Run BlendingEditor.
✅ Check that mapped URL exist in Blender.ListAllByUris()

**Screenshots:**
blend json


src/BlendingEditor/blend.json
```diff
@@ -0,0 +1,9 @@
+[
+  {
+     "Uri": "/blendingeditor/app-name",
+     "Token": "",
+     "Contexts": ["app", "icon"],
+     "AllowFromDirection": true,
```
warpech: Practically this is only used with Launchpad, so I think `AllowFromDirection` can be false

Issue #185, opened 2018-05-02
[Mapping branch] code should be updated from master


kegor:
Continuation for the #182.

kegor:
Closing because related changes were merged in #186.

PR #186, opened 2018-05-02
Problem #185: [Mapping branch] merge with master.

kegor:
Changes for #185.


kegor:
Regarding last failing test:
I recently found that in Firefox code like `RedirectUrl = "/blendingeditor/collections";` doesn't work, I see that URL is changing in the address line, but page content remains the same.

@alshakero suggested changing `url` to `url$` in the lines like `<palindrom-redirect history url$="{{model.RedirectUrl}}"></palindrom-redirect>` but it didn't help for the moment.

kegor:
For discussing the first issue, I had connected with @chrhol, he suggested that the problem might be because of different sessions handles in the AttachedScope.Transaction and in the Transaction.Current, I tried to look at these values right after clicking `Delete` in the grids on the mentioned pages:
image
— this is the case in wich we are opening the page through the link from another page: different Sessions handles.
image
— this is the case where we are using refresh (F5) or opening the page in the blank tab.
This detail can answer the question "Why Session is not dirty in the first case".
Then we discussing about the using `Db.Scope` as the main possible reason of this wrong behavior, and I found only one using:
https://github.com/Starcounter/Blending/blob/2abb1a0c59dd9886c2a0d879999d77a42b3a05b3/src/BlendingEditor/Website/Api/MainHandlers.cs#L136
— but on this point, I was stuck, because I didn't find the solutions for trying to work around somehow.

For the second bug I found that Starcounter-include's list is not populated in CompositionEditor because some of them have `elem.viewModel[elem.compositionProvider]` == null here:
https://github.com/Starcounter/starcounter-layout-html-editor/blob/f8fccbe88fccebcd2367b17e72e654273f51199a/starcounter-layout-html-editor.html#L325
I tried to add simple check on null here and it works, but I do not 100% sure that this might be the appropriate fix for the moment. @PatrykSzwer wdyt?

PatrykSzwer:
About first part TBH I know as much as you, maybe you can sync with @chrhol a little more and together find a solution, because I don't really have an idea what could you do next.

About second part I'm not that familiar wtih this CE, maybe @tomalec could provide you the right answer.

tomalec:
> For the second bug I found that Starcounter-include's list is not populated in CompositionEditor because some of them have `elem.viewModel[elem.compositionProvider]` == null here:
https://github.com/Starcounter/starcounter-layout-html-editor/blob/f8fccbe88fccebcd2367b17e72e654273f51199a/starcounter-layout-html-editor.html#L325
I tried to add simple check on null here and it works, but I do not 100% sure that this might be the appropriate fix for the moment. @PatrykSzwer wdyt?

I'm not sure I get 100% of the problem, but if there is no composition provider,for given view, there is no point to edit it in the editor (as it won't work anyway)

habib535:
> Then we discussing about the using Db.Scope as the main possible reason of this wrong behavior, and I found only one using:

Does this functionality work fine on the master branch? May be you need to revisit the Authorization library for how the Db.Scope was implemented there. If each of [the handler](https://github.com/Starcounter/Blending/blob/2abb1a0c59dd9886c2a0d879999d77a42b3a05b3/src/BlendingEditor/Website/Api/MainHandlers.cs#L118-L128) needs to have a separate transaction then you can try the [this](https://docs.starcounter.io/guides/transactions/long-running-transactions#making-sure-to-create-a-new-transaction) solution.

kegor:
@habib535 in the `master` branch I see using `Db.Scope` only in test-helpers.
I'll try to find the implementation of the `Db.Scope` in Authorization library.
And I'll try to use a separate transaction for each handler.

kegor:
Firstly I've tried to use a separate transaction for each AuthorizedHandle but the result was the same.

Then @un-tone pointed that current changes in `mapping` branch are non-compliance with design:
The main problem is that `partial handlers` with attributes like this https://github.com/Starcounter/Blending/blob/a771fb55152c39773bfe43aa65372fb2586bc3a5/src/BlendingEditor/Website/ViewModels/LayoutGroupsPage.json.cs#L12 were removed. In other words in the initial design for the `LayoutGroupPage` the following URL should be requested `/blendingeditor/partials/layoutgroups` and then `Db.Scope` should be opened inside it. But now the single instance of `MasterPage` is used in different transaction scopes.

@habib535, we are sure that this is the root cause of the first problem (and actually perhaps for the second) and we are suggesting to revisit this place in `mapping` branch.

habib535:
@kegor Thank you for identifying the root cause, I've made the changes. Can you please verify?

https://github.com/Starcounter/Blending/commit/8579e3676b34f24854062df998767e2e2378e535

kegor:
@habib535, @PatrykSzwer I've merged latest changes from `mapping` branch (from Habib https://github.com/Starcounter/Blending/commit/8579e3676b34f24854062df998767e2e2378e535) into current task branch `mapping-merge-with-master` and checked issues mentioned above https://github.com/Starcounter/Blending/pull/186#pullrequestreview-117991305, and now they are not reproducible.

And because tests were also ok, maybe we could finish this PR?

Issue #139, opened 2018-03-22
Signing in requires manual page refresh


mmnosek:
### Steps to reproduce
1. Run `BlendingEditor`
1. Run `SignIn`
1. Open `/BlendingEditor`
1. SignIn with some user

### Expected result
Layout group page should be visible

### Actual result
"Access denied" is visible

##### Starcounter version 
Starcounter-Develop-2.4.0.5497

##### Involved apps' version
* StarcounterApps/Blending@21cecf559285fabd62244251bfeb307f35bbf45a
* StarcounterApps/SignIn@16c9f3ad73ed0abee10ecc561beb90c93b2551c0

PatrykSzwer:
Closing as done. (huurrrayy) 🎆

PR #188, opened 2018-05-04
Refresh partial after user signing in or out.

PatrykSzwer:
For #139


kegor:
I've checked two cases locally:
1. Sign in on the `/blendineditor` page.
2. Sign out on the `/blendineditor` page.
— and they work as expected.
But I also would like to take a look at the code.

PR #189, opened 2018-05-08
Improve condition for picking right composition provider namespace

PatrykSzwer:
For https://github.com/Starcounter/starcounter-layout-html-editor/issues/24
There still will occur some views without Composition property (mostly `Main` partial view) but this is related to viewKeeper and will be presented in another issue.


tomalec:
I think I could be a little bit biased, reviewing my own change ;) code LGTM.

Do you think we could add a test for it?

src/BlendingProvider/Composition/Api/Middleware.cs
```diff
@@ -22,7 +22,16 @@ public static void Register()
     if (typeof constructor !== 'undefined') {
         Object.defineProperty(constructor.prototype, 'compositionProvider', {
             get: function() {
-                return this.partial.BlendingProvider_0 && (typeof this.partial.BlendingProvider_0.Composition$ === 'undefined') && 'BlendingProvider_1' || 'BlendingProvider_0';
+                const blendingProviderKeys = Object.keys(this.partial).filter(key => key.startsWith('BlendingProvider_'));
+                const keysWithComposition = blendingProviderKeys.filter(key =>  this.partial[key].Composition$ !== undefined);
+                debugger;
```
tomalec: remove this debugger

src/BlendingProvider/Composition/Api/Middleware.cs
```diff
@@ -22,7 +22,16 @@ public static void Register()
     if (typeof constructor !== 'undefined') {
         Object.defineProperty(constructor.prototype, 'compositionProvider', {
             get: function() {
-                return this.partial.BlendingProvider_0 && (typeof this.partial.BlendingProvider_0.Composition$ === 'undefined') && 'BlendingProvider_1' || 'BlendingProvider_0';
+                const blendingProviderKeys = Object.keys(this.partial).filter(key => key.startsWith('BlendingProvider_'));
```
tomalec: This delegates the Blending app responsibility to know what itself is doing on the server-side, to customer's browser to figure it out over and over for every partial.

tomalec:
As already stated at https://github.com/Starcounter/Blending/issues/30#issuecomment-369183983

I'm totally against such solutions. It's not only frontend fixing backend bugs.

This delegates the Blending app responsibility to know what itself is doing on the server-side, to customer's browser to reverse-engineer and figure it out over and over for every partial.

src/BlendingProvider/Composition/Api/Middleware.cs
PatrykSzwer: Right now it is the best way to provide namespace with Composition in an unexpected amount of namespaces in case of more apps than only Blending app.

src/BlendingProvider/Composition/Api/Middleware.cs
PatrykSzwer: Ough you got me there, I was already removing it when you commented :p

kegor:
I've checked changes locally, I see this result:
image
— I see that now all blending points have `viewHtml` property except of `blendingeditor/Main`, but if I understand correctly this last item will be fixed in #190? Am I right @PatrykSzwer?

PatrykSzwer:
As I said, this is due to viewKeeper bug. If you check the main in debug tools you will see it refers to  ViewKeeper.html

src/BlendingProvider/Composition/Api/Middleware.cs
```diff
@@ -13,16 +13,21 @@ public static void Register()
...
-                return this.partial.BlendingProvider_0 && this.partial.BlendingProvider_0.Sections && 'BlendingProvider_1' || 'BlendingProvider_0';
+                const blendingProviderKeys = Object.keys(this.partial).filter(key => key.startsWith('BlendingProvider_'));
+                const keysWithComposition = blendingProviderKeys.filter(key =>  this.partial[key].Composition$ !== undefined);
+
+                if(keysWithComposition.length === 0)
+                {
+                    return 'BlendingProvider_0'; // To handle viewKeeper bug
+                }
+
+                return keysWithComposition[0];
```
tomalec: Correct me if, I'm wrong but
1. `LayoutInfo` is the only response added via `JsonResponseMerger.RegisterMergeCallback`,
2. `JsonResponseMerger.RegisterMergeCallback` adds siblings **after** or other regularly blended view-models => they get highest numbers
3. 1&2=> nasz composition provider jest zawsze w ostatnim z `BlendingProvider_*`,
4. BlendingProvider responsów  ma dokładnie 3,

PatrykSzwer: Tested this with multiple apps, seems like there is no case where we crossing this number 2. I'll add this change.

tomalec:
I hope, we will find time to clean up the app and solve it in a way that is as bulletproof as re-discovering composition provider with loops, and cheap as a fixed prop.

PatrykSzwer:
I have to investigate tests a little before merge, because compositionProvider tests are failing right now.

tomalec:
Could you explain what was the problem, and why it's better now? (and update the comment that was explaining the previous condition)

src/BlendingProvider/Composition/Api/Middleware.cs
```diff
@@ -24,7 +24,7 @@ public static void Register()
     if (typeof constructor !== 'undefined') {
         Object.defineProperty(constructor.prototype, 'compositionProvider', {
             get: function() {
-                return this.partial.BlendingProvider_2 &&  'BlendingProvider_2' || this.partial.BlendingProvider_1 && 'BlendingProvider_1' || 'BlendingProvider_0';
+                return this.partial.BlendingProvider_0 && (typeof this.partial.BlendingProvider_0.Composition$ !== 'undefined') && 'BlendingProvider_0' || this.partial.BlendingProvider_2 && 'BlendingProvider_2' || 'BlendingProvider_1';
```
tomalec: Why is that picking `BlendingProvider_0` instead of `BlendingProvider_2`? and why BlendingProvider_last is there and does not contain composition?

PatrykSzwer:
I added more explicit comment to explain current approach to condition about picking right namespace. Those will help remember what was the reason behind this order of namespaces.
I tested this version with multiple apps and tests themselves passed in multiple tries, local and in TC. Due to that, I'll merge this PR, since fix is fairly stable, and this subject will be reconsidered with @warpech when he will be back (and probably with little involvement of @un-tone).

Issue #108, opened 2018-03-06
Improve live preview of layout collections variables changes


PatrykSzwer:
Right now you can't see any reaction during making changes to collection variables unless you click "Save" button. Eventually, we want user to be able to see his changes right after writing a new value to every variable.

kegor:
Just thoughts aloud: the initial idea can be continued to this checkbox:
![image](https://user-images.githubusercontent.com/17431807/37774086-4ed53346-2df0-11e8-8023-e6fd216a3b59.png)

kegor:
Theoretically, we can achieve this by following steps:
1. In BE CollectionPage handler we can get Collection/CssVariable names and changed value.
2. Then using the Blender Mapping we can pass these parameters into the BP handler.
3. In the separate method, we can recalculate CssVariables string and replace the old value in each LayoutPage.CssVariables which can be found in current session.
Probably the main side effect of this approach is that if a user leaves tab after changing CSS variable value without saving then all current changes are left on the page. And seems like we need yet another handler to revert changes back.

So maybe make sense to try to implement this feature on a client-side.

kegor:
@warpech, @un-tone what do you think?

un-tone:
At first glance, it seemed over-complicated to me if using habitual approaches here, I mean all work on server side, especially considering the requirement to revert changes back if a user goes away, cancels changes or maybe does something else.
But probably implementing on a client side will not be easier because it will also require implementing reverting the CSS styles, it is unclear to me how it can be done because the client side should be in sync with the server side.

Probably, it is possible to make some changes in and improve the CSS Variables implementation to make it more convenient to use and manipulate with its data. Currently, it looks quite hard to do so. Just my point of view, I could miss something.

warpech:
Regarding the problem in the OR, it seems to me that we already have the server and client solution, just for some reason the transaction is isolated until the `Save` button is pressed on the CSS variables editing page.

The two-way data binding between the view-model and the view is already defined here: https://github.com/Starcounter/Blending/blob/4486d941cd5ed6ffc6c83fa69b456eb4e13cef66/src/BlendingProvider/Website/Api/ContentHandlers.cs#L75

The view-model uses the getter: https://github.com/Starcounter/Blending/blob/4486d941cd5ed6ffc6c83fa69b456eb4e13cef66/src/BlendingProvider/Website/ViewModels/LayoutPage.json.cs#L15. This getter re-executes each time a change comes from the client.

---

[...]

PR #193, opened 2018-05-10
Problem #108: CssVariables live preview.

kegor:
**Problem #108:**
>Right now you can't see any reaction during making changes to collection variables unless you click "Save" button. Eventually, we want user to be able to see his changes right after writing a new value to every variable.

**Solution:**
In the initial idea for work-aroung I'm trying to commit transaction after each change of the `CssVariable.Value` and then make `redirect` on the same page of the `CssCollection`.

**Screenshots:**
css-variables-live-preview


PatrykSzwer:
Important information related to possible next steps/solution: https://github.com/Starcounter/Blending/issues/108#issuecomment-388403291

warpech:
Closing this as we are pursuing another solution in https://github.com/Starcounter/Blending/issues/108#issuecomment-388403291

Thanks @kegor, @PatrykSzwer

Issue #52, opened 2018-01-15
Add themability of Paper and Vaadin components


warpech:
After https://github.com/Starcounter/Blending/issues/51 is done, see if we can go any further by allowing to theme Paper and Vaadin web components.

warpech:
Depends on https://github.com/Starcounter/RebelsLounge/issues/258.

This issue will be about adding some UI or preconfiguration for it.

PatrykSzwer:
Closing as done in #199

warpech:
Note to future self. This feature will be removed. See: https://github.com/Starcounter/Blending/issues/240

PR #199, opened 2018-05-17
Themability feature

PatrykSzwer:
For #52 

This feature is based on adding `<enlighted-link>` with path to the theme into HTML composition. (I also changed a little how transactions works at general page to be more intuitive.)


tomalec:
How can I test it?
I tried:
1. Running Blending, SignIn, KitchenSink,
2. Go to http://localhost:8080/BlendingEditor/layoutgroups
3. Pick "TopbarLayout",
4. Click Manage themes,
5. Save "vaadin-material","	https://raw.githubusercontent.com/vaadin/vaadin-material-theme/master/vaadin-date-picker.html",
6. Reload page,
7. Go to http://localhost:8080/KitchenSink/Datepicker 
8. Reload page

The theme is not applied also I cannot see any new `<enlighted-link>` in starcounter-include[slot="blendingprovider/content"]

themes-manager2

PatrykSzwer:
TBH I did pretty much exactly the same steps as you did and everything works for me. GIF with steps:
repro

My env:
Starcounter 2.4.0.6130
this branch of Blending app (obv)
SignIn
Newest master-2.4 kitchensink.

Maybe you need to update SC? Let me know if you find out the reason.

tomalec:
Added more steps to reproduce to https://github.com/Starcounter/Blending/pull/199#issuecomment-390172105), to make it reproducible on Patryk's machine. We need to edit composition before adding a theme to it.

For some reason, edited composition is not reflected in `Starcounter.HTMLViewComposition`.

PatrykSzwer:
@tomalec I pushed fix for our issue. 
I talked with @un-tone and turns out that bug in starcounter administrator mislead us. (too big entities wasn't showing up). 
Compositions works with mechanism of revisions. Every change in composition is saved as new row in the database with special pointer (NextVersion column) in the old one, pointing to the new one. So the solution was to look for a composition with unset (null) NextVersion column, which indicates that this is the newest, current version.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -163,18 +175,92 @@ void Handle(Input.SaveModalTrigger action)
...
+                // Current version of composition has NextVersion value set to NULL, rest of them are previous revisions
+                var composition =
+                    Db.SQL<HtmlViewComposition>(
+                        $"SELECT c FROM {typeof(HtmlViewComposition)} c WHERE c.{nameof(HtmlViewComposition.MainView)}.Uri = ? " +
+                        $"AND c.{nameof(HtmlViewComposition.NextVersion)} IS NULL",
+                        this.Data.Html).FirstOrDefault();
+                if (composition != null && !string.IsNullOrEmpty(this.ThemeName) && !string.IsNullOrEmpty(this.ThemePath))
+                {
+                    var enlightedLink = $"<enlighted-link rel='import' name='{this.ThemeName}' href='{this.ThemePath}'></enlighted-link>\n";
+                    composition.Html = enlightedLink + composition.Html;
```
un-tone: You should not change existed compositions, instead create a new version for it and make changes there. Please fix here and in other such places.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -163,18 +175,92 @@ void Handle(Input.SaveModalTrigger action)
...
+
+            if (string.IsNullOrEmpty(composition?.Html))
+            {
+                throw new NullReferenceException("Composition for current layout couldn't be found!");
+            }
+
+            return composition;
+        }
+
+        private void GetCurrentThemes()
```
un-tone: I guess it should be named as `LoadCurrentThemes` or `Refresh...` since it returns nothing

PatrykSzwer: Nice catch, totally agree.

tomalec:
Now, my scenario is:

1. Running Blending, SignIn, KitchenSink (you can reset the data, but don't have tp),
2. Go to http://localhost:8080/BlendingEditor/layoutgroups
3. Pick "TopbarLayout",
4. Click Manage themes,
5. Save "vaadin-material","	https://raw.githubusercontent.com/vaadin/vaadin-material-theme/master/vaadin-date-picker.html",
1. Go to http://localhost:8080/KitchenSink/Datepicker
1. Edit `starcounter-include[slot="blendingprovider/content"]` composition by adding `foo` at the top,
8. Reload page (theme and `"foo"` are there)
9. Go back to http://localhost:8080/BlendingEditor/layout/{id}/general

## Expected result

The theme is still there

## Actual result

Themes card is empty, even though theme is applied in the preview.
theme_is_gone

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -163,18 +177,92 @@ void Handle(Input.SaveModalTrigger action)
...
+
+        private void LoadCurrentThemes()
+        {
+            var compositionHtml = this.GetLayoutComposition().Html;
+            if (string.IsNullOrEmpty(compositionHtml))
+            {
+                throw new NullReferenceException("Composition for current layout couldn't be found!");
+            }
+
+            var pattern = @"<enlighted-link rel='import' name=['|""]([\w| |\-|_]+)['|""] href=['|""](.+)['|""]></enlighted-link>"; //Find every enlighted-link in composition, and get name with path from it
```
tomalec: What If I add an additional attribute whitespace, or switch the order, that it makes it no longer a "theme"?

PatrykSzwer: Yes, It's not designed to be manually edited.

tomalec:
https://github.com/Starcounter/Blending/pull/199#discussion_r189875770
> Yes, It's not designed to be manually edited.

I understand that this means, that it's designed that UI in theme manager will only read/support `<enlighted-link>`s created by itself/in very strict format, and may not support when the element instance is changed. To me it sounds slightly limmiting, but acceptable.

But what about being able to change themed view composition without touching `<enlighted-link>` instance created by Theme manager?
https://github.com/Starcounter/Blending/pull/199#issuecomment-390961692

PatrykSzwer:
I'm checking your scenario right now.

PatrykSzwer:
>Themes card is empty, even though theme is applied in the preview.

For some reason everything is fine except regex which is not catching enlighted-link as a match anymore. I'll fix that.

PatrykSzwer:
@tomalec turns out that there was a small bug in the regex pattern. It should work correctly now.

tomalec:
Scenario above works :+1:

But I have another one:
1. Create themes:
 - `foo, bar`
 - `bar, baz`
 - `foo, baz`
2. Remove one of `foo`s

### Expected
Should be removed

### Actual
Is still there

deleting theme

PatrykSzwer:
TBO I can't reproduce that. Please make sure you have everything setup correctly (maybe you have some custom change in composition which causes the issue?) and/or provide more specific reproduction steps.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -163,18 +177,92 @@ void Handle(Input.SaveModalTrigger action)
...
+                throw new NullReferenceException("Composition for current layout couldn't be found!");
+            }
+
+            return composition;
+        }
+
+        private void LoadCurrentThemes()
+        {
+            var compositionHtml = this.GetLayoutComposition().Html;
+            if (string.IsNullOrEmpty(compositionHtml))
```
un-tone: This is redundant since you already have such checking in the `GetLayoutComposition` method.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -163,18 +177,92 @@ void Handle(Input.SaveModalTrigger action)
...
+
+        [GeneralPage_json.Themes]
+        partial class GeneralThemePage : Json
+        {
+            private GeneralPage ParentPage => this.Parent.Parent as GeneralPage;
+
+            public void Handle(Input.RemoveThemeTrigger action)
+            {
+                ParentPage.RemoveThemeFromComposition(this.Name, this.Path);
+                ParentPage.Themes.Remove(this);
```
un-tone: It seems too many actions from a child element scope. How about moving `ParentPage.Themes.Remove(this);` to `RemoveThemeFromComposition` method and change its name to `RemoveTheme`?
Also, I wonder why you don't use the approach [with delegates](https://docs.starcounter.io/v/2.3.0/cookbook/accessing-methods-in-nested-view-models) instead of explicit retrieving and casting `this.Parent.Parent`?

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -163,18 +177,92 @@ void Handle(Input.SaveModalTrigger action)
...
+                if (composition != null && !string.IsNullOrEmpty(this.ThemeName) && !string.IsNullOrEmpty(this.ThemePath))
+                {
+                    var enlightedLink = $"<enlighted-link rel='import' name='{this.ThemeName}' href='{this.ThemePath}'></enlighted-link>\n";
+                    DataHelper.GenerateNewComposition(composition, enlightedLink + composition.Html);
+                }
+
+                this.ThemeName = string.Empty;
+                this.ThemePath = string.Empty;
+
+                this.LoadCurrentThemes();
```
un-tone: I guess it is not needed to call `LoadCurrentThemes` if `ThemeName` or `ThemePath` was empty, right?
Also maybe better to not close the modal window in this case?

PatrykSzwer: I was covering the situation when someone is deleting theme and then clicking save button, but now I see that It will be covered anyway by the delete method of a theme. About the second part I actually like the way how is it working right now about closing modal, but we will have another round of changing UI for it anyway, so this can be changed in next iteration.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
PatrykSzwer: I changed the method to removeTheme. About changing the approach to the delegate one, I found the one with the Parent page more convenient in this case, and clearer/shorter.

tomalec:
### Steps

1. Create themes:
 - `foo, bar`
 - `bar, baz`
2. Open composition editor,
3. Remove newline after first `<enlighted-link>`
4. Open theme manager

### Expected
foo and bar themes should still be there

### Actual
Only `foo` is detected with a corrupted path

lines

PatrykSzwer:
Me and @alshakero changed approach for the finding/removing part a bit. Now It's not regex based but XML parse based. It should be more bulletproof and save solution. @tomalec please check.

Issue #146, opened 2018-03-23
Open Composition Editor in a new window


mmnosek:
Having `CompositionEditor` above the website is very inconvenient because:
* it takes space on my screen and makes the actual application smaller
* sometimes if there are some full-screen-size elements with `z-index` it's hard to use it at all
* editor is small, so in a little bit more complex compositions working with this is quite hard

I wanted to edit this:
![blendingeditor_and_windows_10](https://user-images.githubusercontent.com/4943949/37829071-e736c5f2-2e9d-11e8-8cba-2e4421c3743c.png)

And I end up with this:
![blendingeditor](https://user-images.githubusercontent.com/4943949/37829132-23cae8f4-2e9e-11e8-9aff-9c2b9b0a3001.png)

I think we should move it to the additional window, if possible. 

@warpech @tomalec WDYT? Low prio.

alshakero:
PR in #205

Issue #24, opened 2018-01-08
Display the key for the composition in the database


PatrykSzwer:
_Issue moved from CompositionEditor repository: https://github.com/Starcounter/CompositionEditor/issues/31_

When editing the composition, I would like to know the [key](https://github.com/Starcounter/level1/blob/2d6990034e47906453c8108c6191bdf43aa78959/src/Starcounter.Extensions/HTMLComposition.cs#L8) by which it will be saved in the database. 

This way it becomes easier to make notes for exporting/importing of the saved compositions.

Right now the way to do it is to use Chrome Dev Tools Network tab to check the REST API request.

![artboard](https://user-images.githubusercontent.com/566463/28121580-dc624fe2-671b-11e7-91d5-0ad89ada74bc.png)

un-tone:
The issue seems to be outdated since `HTMLComposition` is not used anymore and the current data model for compositions does not contain the `Key` property or any alternative. It means that the data model does not have any representative properties which would be useful in the Composition Editor view.

I see two ways:

1. Close the issue as irrelevant.
2. Add a new property to the data model in level1.

kegor:
Thank you @un-tone for clarification, @warpech, @PatrykSzwer wdyt?

PatrykSzwer:
Depends, if we still think that this feature is needed and will be used by users. @warpech was the initiator of this idea so I'll let him make the decision.

warpech:
The purpose of this change is to remove the need of using Starcounter DevTools to find out what are the views that take part in this particular composition and where it is stored in the db.

This issue is not obsolete, but since we changed the data model, the message needs to be adapted. I believe it should be something like below.

Most information should be already in the JSON, if I recall correctly only the object ID of `HtmlViewComposition` should be added to it.

![new](https://user-images.githubusercontent.com/566463/40312392-106409d2-5d13-11e8-9d2e-d8081c95f7c0.png)

alshakero:
Closing as done in https://github.com/Starcounter/Blending/pull/205 & https://github.com/Starcounter/Blending/pull/244.

Issue #149, opened 2018-03-26
Implement simple composition editor


warpech:
Apart from the HTML composition editor, this app needs a demo-friendly simple editor that has just a few options:

- move slotted elements up and down, maybe also left and right
- group and ungroup

This is yet to be fully defined. Nothing more than a useful **minimum** is needed.

tomalec:
Should it be done instead of our current tree-like, visual editor, or as a third option?

warpech:
As a third option.

warpech:
It was discussed by @tomalec and me today, that the initial MVP of the simple editor will be part of a floating toolbar (#22).

Simple editor will have the following options:

1. Select direct child elements of shadow root by point and click, using the same highlight mechanism that is used to select blending points in #22.
   - for many Uniform views, this will only select `div.uni-section`, but that is fine
2. Buttons to move the selected elements up/down in the DOM tree (also if the element is `<slot name="...">`)
3. Button to hide the selected element by putting a `hidden` attribute on it (also if the element is `<slot name="...">`)
4. Button to toggle showing of hidden elements, that allows to unhide them
5. Control buttons that are the same as in the advanced editor: Save, Cancel, Restore default

PR #205, opened 2018-05-24
Make composition editor run in a popup, show object ID and views URLs

alshakero:
The changes of `starcounter-layout-html-editor` are included in this PR for your comfort while testing. They're not meant to be reviewed as code, but just to let you run the full experience without cloning several repos. Once you approve I'll remove the changes and `bower install` the to-be released version of `starcounter-layout-html-editor`.

To review `starcounter-layout-html-editor` changes please review https://github.com/Starcounter/starcounter-layout-html-editor/pull/33 first, once you agree with the code there, you can review here and ignore `starcounter-layout-html-editor` files since you already reviewed them in 33.

Fixes https://github.com/Starcounter/Blending/issues/146 and https://github.com/Starcounter/Blending/issues/24


src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -1,68 +1,60 @@
...
-                var panel = document.querySelector("compositioneditor-panel");
-                if (panel) {
-                    panel.parentNode.removeChild(panel);
-                    panel.removeEventListener('close', closePanel);
-                    setInsertersState(false);
-                }
+                /* we need to globalize this, so the popup can connect to it.
+                 we could say popup.juicyHighlightInstance =  juicyHighlightInstance instead,
+                 but we need the popup to continue working even after refreshing it. This method allows
+                 the popup to always do `opener.juicyHighlightInstance`. Which will stay defined even
```
tomalec: Should it be the same instance after refresh? Refresh of which page, popup or opener?

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -1,68 +1,60 @@
...
 
-            function closePanel() {
-                var panel = document.querySelector("compositioneditor-panel");
-                if (panel) {
-                    panel.parentNode.removeChild(panel);
-                    panel.removeEventListener('close', closePanel);
-                    setInsertersState(false);
-                }
+                /* we need to globalize this, so the popup can connect to it.
+                 we could say popup.juicyHighlightInstance =  juicyHighlightInstance instead,
```
tomalec: Why can't we do `opener.querySelector('juicy-highlight[compositioneditor-inserter]')`? instead of storing global?

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -1,68 +1,60 @@
...
+                 but we need the popup to continue working even after refreshing it. This method allows
+                 the popup to always do `opener.juicyHighlightInstance`. Which will stay defined even
+                 after refresh */
+                window.juicyHighlightInstance = juicyHighlight;
             }
-
             Polymer({
                 is: "compositioneditor-inserter",
-                properties: {
-                    active: {
```
tomalec: Are you sure we no longer use `active` property to bind some styling somewhere?

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -1,68 +1,60 @@
...
-                    if (this.active) {
-                        showPanel();
-                    } else {
-                        closePanel();
-                    }
+                _showPanel() {
+                    popup = window.open('/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-popup.html', 'CompositionEditorPopup', 'toolbar=no, location=no, directories=no, status=no, menubar=no, scrollbars=no, resizable=no, copyhistory=no, width=1200, height=500');
+
+                    /* Same as above, we need to globalize this, so the popup can connect to it even after refresh. */
+                    window.whereToHighlight = this;
```
tomalec: It is polluting, and I don't get why it's needed. Could you elaborate on what refresh, why it would be lost etc? Sorry maybe I don't get it because of my illness 🤒

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-popup.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+    <meta charset="utf-8">
+    <title>Composition Editor</title>
+
+    <script src="/sys/webcomponentsjs/webcomponents-lite.js"></script>
+    <link rel="import" href="/sys/polymer/polymer.html">
```
tomalec: I don't think we use Polymer in this document. I'd remove it.

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-panel/compositioneditor-panel.html
```diff
@@ -1,94 +0,0 @@
...
-        }
-    </style>
-    <enlighted-link rel="import" href="/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html"></enlighted-link>
-    <!--
-    Load ace/mode/html manually
-    This is a workaround for a conflict between ace loaded by juicy-ace-editor
-    and browserified (br)ace loaded by jsoneditor (juicy-jsoneditor)
-    https://github.com/josdejong/jsoneditor/issues/309
-    -->
-    <script src="/sys/ace-builds/src-noconflict/mode-html.js"></script>
```
tomalec: Don't we need this hack any longer in the popup page?

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
alshakero: Pop up.

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
alshakero: Hmmm sounds like a great idea. I'll check it.

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
alshakero: Your suggestion above also applies here I guess. A querySelector should do. I was so invested in getting it to work.

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-panel/compositioneditor-panel.html
alshakero: Not really. Because it's in another window. They won't affect each other anymore.

tomalec: That's why I asked, why it's not added to `...popup.html`

tomalec:
Besides Polymer, there is something wrong with the overlay:
1. Open editor,
2. Select something from dropdown,
3. Refresh editor popup,
4. Close the editor,

### Expected
The overlay should be gone.

### Actual 
The overlay is still there.

overlay

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-panel/compositioneditor-panel.html
alshakero: I mean we don't need this hack any more because popup.html and the host window are seperate. So `ace-editor` instances will to see each other at all. Am I missing something?

alshakero:
> Besides Polymer, there is something wrong with the overlay:

Are you sure you pulled? On my machine the overlay is disappearing at the moment I refresh.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -18,10 +17,10 @@
          <style>
             :host{
                 display: flex;
-                resize: both;
                 overflow: auto;
 
                 flex-direction: column;
+                height: calc(100% - 50px);
```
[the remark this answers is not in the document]
alshakero: We don't need this, you're right. Eariler when I made it 100% it overflew and the buttons disappeared. Maybe giving the `body` `box-sizing: border-box` eliminated this need.

tomalec:
> Are you sure you pulled? 

I was, but after I pulled also Polymer fix and reinstalled SC, the problem is gone.

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-popup.html
```diff
@@ -0,0 +1,23 @@
...
+    <title>Composition Editor</title>
+
+    <script src="/sys/webcomponentsjs/webcomponents-lite.js"></script>
+    <link rel="import" href="/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html" />
+    <style>
+        html, body {
+            height: 100%;
+            margin: 0;
+            box-sizing: border-box;
+            padding: 5px
```
tomalec: Could you add semicolon here, to avoid accidental bug when a rule is added bellow?

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -296,30 +310,41 @@
...
                 // Fires when the local DOM has been fully prepared
                 ready: function() {
+                    this.juicyHighlightInstance = createAndInjectJuicyHighlight();
                     this.addEventListener('mouseup', function(ev) {
-                        window.dispatchEvent(new Event('resize')); //fix resizing of ACE edior
+                        opener.dispatchEvent(new Event('resize')); // fix resizing of ACE edior
                     });
                     this.refreshSelectors();
+
+                    // we don't need to remove those event listeners, the window is disposed right after they're dispatched
```
tomalec: But when the element is removed from DOM, we probably don't want to listen to them any longer.

alshakero: The element will never be removed the window, because the window is the popup in this context.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -214,7 +227,8 @@
                     if (this.partial) {
                         // console.log('_selectorChanged to smth', this.partial);
                         this.set('disabled', false);
-                        this.$.highlightScopeSelected.show([this.parentElement, this.partial]);
+                        const whereToHighlight = opener.document.querySelector('compositioneditor-inserter');
+                        this.juicyHighlightInstance.show([whereToHighlight, this.partial]);
```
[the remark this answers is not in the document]
alshakero: I'm not sure I understand correctly. Is this correct https://github.com/Starcounter/Blending/pull/205/commits/571abc2a9b7bb235e031c5b1c24e02eca96caea7?

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
tomalec: Any element could be removed from any tree from any window, that just a matter of well behaving custom element.

Having it as a custom element implies DOM API, therefore `starcounterLayoutHTMLEditor.remove()` is a part of API even if we are not using it yet.

tomalec: I'd remove listeners on `disconnectedCallback` to avoid hard to track problems in the future.

alshakero: Solid points 😞 

haha

alshakero:
I'd really appreciate it if we could finish this before the weekend.

alshakero:
Could you approve https://github.com/Starcounter/starcounter-layout-html-editor/pull/33 I could merge and release?

tomalec:
What about https://github.com/Starcounter/starcounter-layout-html-editor/pull/33/files#r190826818  and https://github.com/Starcounter/starcounter-layout-html-editor/pull/33#discussion_r191209755?

alshakero:
> What about Starcounter/starcounter-layout-html-editor/pull/33/files#r190826818 and Starcounter/starcounter-layout-html-editor#33 (comment)?

Both are addressed

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -138,7 +163,37 @@
...
+            <span>Changes will be saved in the table <code>Starcounter.HtmlViewComposition</code></span>
+             
+            <dom-if if="{{CompositionKey}}">
+                <template>
+                    | Object ID: <strong><code>{{CompositionKey}}</code></strong> 
+                </template>
+            </dom-if>
+            <dom-if if="{{isEmpty(CompositionKey)}}">
+                <template>
+                    | No Object ID [<a href="javascript:;" title="Default compositions they don't have object IDs, as they aren't stored in the database">?</a>]
```
tomalec: Maybe something like

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -15,12 +15,22 @@
...
                 height: 100%;
+               
+                padding: 10px;
+                box-sizing: border-box;            
+            } 
+            *:not(code) {
+                /* Windows is the lead clientele */
+                font-family: 'Segoe UI', Tahoma, Geneva, Verdana, sans-serif;
+                
+                color: #444;
```
tomalec: Is this 
image a desired behavior? Personally, I find this _`id`_ with different font and font color slightly strange. I used  this `<em>`phasis to suggest that `id` is a variable string, so I could expect different color and italics, but changing the font from code'like monotype for something meant for `:not(code)` is strange. I believe `*:is(:not(code))` would be desired, not to increase specificity.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -219,8 +274,24 @@
...
+                },
                 observers: [
                 ],
+                _getViewsFromTemplate(template) {
+                    // first one is the main
+                    const scope = template.scopedNodes[0].scope;
+                    const mainView = this.partial.viewModel[scope].Html;
+
+                    // the rest is attached nodes
+                    const attachedViews =  template.scopedNodes.slice(1).reduce((accumulator, scope) => {
```
tomalec: Am I correct that, you are trying to reverse engineer the set of attached views here?

tomalec: If so, I'd say:
 - we should not reverse engineer ourselves, and return it from the server-side ;),
 - if we really have no other choice:
   - it would be nice to mention why we are doing so,
   - we should use view-model, not view to do that.
     There could be a case where there is an attached view, that should participate in this array, but the view returns no nodes to be stamped (yet, there is just set of dependencies/themes, etc.) so `template.scopedNodes` could be empty

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
warpech: What's `<a href="javascript:;"`?

warpech: I would propose to write it a more user friendly way:

> This is a default composition that will be stored in the database if you make any changes to it and press "Save".

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -199,7 +254,7 @@
                     },
                     toggleWYSIWYG: {
                         type: Boolean,
-                        value: true
+                        value: false
```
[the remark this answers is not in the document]
warpech: It is beyond the scope of this PR to change that value to `false`.

-----

Btw. @tomalec you refer to as WYWIWYG here is not really a WYSIWYG editor, but a tree-view representation of DOM. 

WYSIWYG HTML editor looks like that:

at in new sites

Jocke wants to have a simple editor enabled by default (https://github.com/Starcounter/Blending/issues/149), but so far we do not have that.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -206,34 +274,62 @@
...
                         this.$.ace.value = this.partial.shadowRoot.innerHTML || '';
                         this._compositionIsBeingSet = false;
-                        // debugger
                         this.$.wysiwyg.doc = this.partial.shadowRoot;
+
+                        const CompositionKey = this.partial.viewModel[this.partial.compositionProvider].CompositionKey;
+
+                        this.set('CompositionKey', CompositionKey);                        
+
+                        if(this.partial.template.scopedNodes[0]) {
```
warpech: It should rather be:

```js
this.set('MainView', this.partial.viewModel.MainView);
this.set('AttachedViews', this.partial.viewModel.AttachedViews);
```

without all those ifs. See: https://github.com/Starcounter/Blending/pull/205/files#r191602208

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
warpech: > we should not reverse engineer ourselves, and return it from the server-side ;)

That's what I said as well. Please let us not do workarounds as our default mode of work.

I think @kegor can help to populate MainView and AttachedViews on the server side.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: > What's <a href="javascript:;"

This is to make the element focusable on screen readers and make it show a pointer cursor without anyside effects when clicking it. Anchors with `href='#'` scrolls to the top when clicked.

> This is a default composition that will be stored in the database if you make any changes to it and press "Save".

I like this one too.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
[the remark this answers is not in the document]
alshakero: Changed it back to `true`.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: Fixed by moving the font setting to `:host`.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: I would love that!

tomalec:
Maybe we can separate this ID thing into another PR, not to block popup feature from rotting?

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
warpech: > This is to make the element focusable on screen readers and make it show a pointer cursor without anyside effects when clicking it. Anchors with href='#' scroll to the top when clicked.

I'd rather just use `<abbr>`.

Btw the chance of someone using a screen reader to use composition editor are close enough to zero to not worry about screen readers at this point.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
warpech: I will be happy to do that, but please tell me what should be explained given that there already is an explanation in https://docs.starcounter.io/v/2.4/topic-guides/blendable-web-apps/view-model-attaching.

The data model does not look complex:

- https://github.com/Starcounter/level1/blob/922f92920e6198b6f75995003ee2bf5fcd917bc5/src/Starcounter.Extensions/HtmlViewComposition.cs
- https://github.com/Starcounter/level1/blob/922f92920e6198b6f75995003ee2bf5fcd917bc5/src/Starcounter.Extensions/HtmlView.cs
- https://github.com/Starcounter/level1/blob/922f92920e6198b6f75995003ee2bf5fcd917bc5/src/Starcounter.Extensions/HtmlViewAttachment.cs

I will be very happy to explain any detail that I missed.

kegor: I'm working on this.

alshakero:
It seems Egor is working on this. I don't think it will rot.

tomalec:
Something is wrong with scrollability of ace editor, see as it stops before the end:

image

With master it looks like
image

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -8,20 +8,27 @@
...
                 display: flex;
-                resize: both;
                 overflow: auto;
+                background: #fefefe;
 
                 flex-direction: column;
+                height: 100%;
+               
+                padding: 10px;
+                box-sizing: border-box;
```
tomalec: This seems to break scrolling (see screenshots in previous comment)

alshakero:
I can't repro https://github.com/Starcounter/Blending/pull/205#issuecomment-393345892. @PatrykSzwer could you please pull and tell me if you could?

PatrykSzwer:
>I can't repro #205 (comment). @PatrykSzwer could you please pull and tell me if you could?

I confirm, I can't reproduce this. @tomalec please provide some reproduction steps if you still experiencing such problem. 

BTW Omar, referring to our call from couple days, could you set Visual editor checkbox to false by default?

alshakero:
> BTW Omar, referring to our call from couple days, could you set Visual editor checkbox to false by default in this PR?

@warpech thinks (and I agree) that's it's out of the scope of this PR. If it affects your flow please create an issue (and I agree it should go away).

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -327,6 +325,7 @@
                         this.juicyHighlightInstance.hide();
                     }
                     this._updateObserver();
+                    setTimeout(() => window.dispatchEvent(new Event('resize'))); // fix resizing of ACE editor
```
tomalec: Could you check if latest Ace still requires this hack? https://github.com/Juicy/juicy-ace-editor/pull/44

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -206,34 +294,46 @@
...
+                        this.set('mainView', mainView); 
+                        this.set('attachedViews', attachedViews);
+                       
                     } else {
                         this.set('disabled', true);
-                        this.$.highlightScopeSelected.hide();
+                        this.juicyHighlightInstance.hide();
                     }
                     this._updateObserver();
+                    setTimeout(() => window.dispatchEvent(new Event('resize'))); // fix resizing of ACE editor
```
tomalec: Maybe I'm missing something, but I set
```css
            juicy-ace-editor{
                flex: 1 0 5em;
                height: 100%; /*<---*/
            }
```
and it works fine without resize event l.328

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -140,7 +180,40 @@
...
-            </div>
+        </div>            
+        <div id="db-notice">
+            <span>Changes will be saved in the table <code>Starcounter.HtmlViewComposition</code></span>
+             
+            <dom-if if="{{compositionKey}}">
+                <template>
+                    | <em>Object ID: </em><code>{{compositionKey}}</code>
+                </template>
+            </dom-if>
```
warpech: We are in a Polymer element, so that's not the recommended syntax to be used. It should be `<template is="dom-if" if="{{compositionKey}}">`.

See: https://www.polymer-project.org/2.0/docs/devguide/templates#dom-if

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -140,7 +180,40 @@
...
+                <template>
+                    | <em>Main view</em>: <code>{{mainView}}</code>
+                </template>
+            </dom-if>
+            <dom-if if="{{attachedViews.length}}">
+                <template>
+                    <div class="attached-views-wrapper">
+                        <em>Attached views:</em>
+                        <ul class="attached-views">
+                            <dom-repeat items={{attachedViews}}>
```
warpech: It should be `<template is="dom-repeat" items="{{attachedViews}}">`

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -140,7 +180,40 @@
...
+        </div>            
+        <div id="db-notice">
+            <span>Changes will be saved in the table <code>Starcounter.HtmlViewComposition</code></span>
+             
+            <dom-if if="{{compositionKey}}">
+                <template>
+                    | <em>Object ID: </em><code>{{compositionKey}}</code>
+                </template>
+            </dom-if>
+            <dom-if if="{{isEmpty(compositionKey)}}">
```
warpech: Isn't `if="{{!compositionKey}}"` enough? It should work according to https://stackoverflow.com/questions/41631186/polymer-dom-if-how-do-i-implement-a-negative-condition

src/BlendingProvider/Composition/Api/JsonResponseMergerCallback.cs
```diff
@@ -114,13 +114,24 @@ public static Json OnJsonMerge(Request request, string callingAppName, IEnumerab
...
             {
+                // populate the ViewUris array for default compositions
+                foreach (var view in partialsUris)
+                {
+                    layoutInfo.ViewUris.Add().SetStringValue(view);
+                }
+            } 
+            else
+            {
+                // populate the ViewUris array for compositions coming from the database
```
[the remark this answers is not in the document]
alshakero: > nomenclature

🍷

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: It works. But I didn't want it to show when `compositionKey` is `undefined`. But this will never happen, right?

alshakero:
Ready for a round. I hope a final.

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -184,35 +184,25 @@
...
-                                <template>
-                                    <li>{{item}}</li>
-                                </template>
-                            </dom-repeat>
-                        </ul>
-                    </div>
-                </template>
-            </dom-if>        
+            <template is="dom-if" if="{{compositionKey}}">
+                | <em>Object ID: </em><code>{{compositionKey}}</code>
```
warpech: Maybe change to a more friendly `<em>Custom composition ID: </em><code>{{compositionKey}}</code>`

src/BlendingEditor/wwwroot/sys/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -184,35 +184,25 @@
...
+            </template>
+            <template is="dom-if" if="{{!compositionKey}}">
+                | <em>No Object ID</em> [<a href="javascript:;" title="This is a default composition that will be stored in the database if you make any changes to it and press 'Save'.">?</a>]
```
warpech: Maybe change to a more friendly `<em>Default composition</em> [<a href="javascript:;" title="This is a default composition that will be stored in the database if you make any changes to it and press 'Save'.">?</a>`

alshakero:
I addressed everything and copied all the changes there If you could approve it https://github.com/Starcounter/starcounter-layout-html-editor/pull/34 

I'll release `starcounter-layout-html-editor` and bower update it here.

alshakero:
Finally done!

Issue #204, opened 2018-05-23
Replace SQL strings with Starcounter.Linq


PatrykSzwer:
@miyconst pointed me out that in Blending app we're using SQL string everywhere, e.g
```
this.Areas.Data = Db.SQL<WebSection>($"SELECT s FROM {typeof(WebSection).FullName} s " +
                                                          $"WHERE s.{nameof(WebSection.Template)} = ? " +
                                                          $"AND s.{nameof(WebSection.Deleted)} = ? " +
                                                          $"ORDER BY s.{nameof(WebSection.Template)}.{nameof(WebTemplate.Name)}, s.{nameof(WebSection.Name)}",
                                                          this.Data, false);
```

Currently, there is a better and cleaner way to achieve that with use of `Starcounter.Linq`: https://github.com/Starcounter/Starcounter.Linq (Nuget package)
I really like the API of that and I think we should replace all our database queries with those functions.

Issue #181, opened 2018-04-26
[Mapping branch] Import feature does not work


warpech:
Leftover from https://github.com/Starcounter/RebelsLounge/issues/240.

@kegor wrote:

Import data on `Settings` page is not working with exception.

warpech:
@kegor wrote:

after [I added](https://github.com/Starcounter/Blending/commit/bfebe6eb132afc1e6871b36a5c06920f28669ad8) `--mapping-mode=true` flags into `unload/reload` StarDump commands I see this error message:
```
 Microsoft.Extensions.CommandLineUtils.CommandParsingException: Unexpected value 'true' for option 'mapping-mode'
 at Microsoft.Extensions.CommandLineUtils.CommandLineApplication.Execute(String[] args)
 at StarDump.CommandLineInterface.Execute(String[] args) in C:\TeamCity\BuildAgent\work\sc-17110\StarDump\src\StarDump\CommandLineInterface.cs:line 36
 
 C:\Program Files\Starcounter\StarDump>
```
Starcounter version: 2.4.990.71
@miyconst, could you please clarify: I'm using mapping Starcounter with the old StarDump version? @habib535 said that you could help with this question.

warpech:
@miyconst wrote:

>Regarding the last issue from the list above, after [I added](https://github.com/Starcounter/Blending/commit/bfebe6eb132afc1e6871b36a5c06920f28669ad8) `--mapping-mode=true` flags into `unload/reload` StarDump commands I see this error message:

The flag does not need any value, just add it as `--mapping-mode`.

warpech:
@kegor wrote:

@miyconst, thank you for this clarification, without `=true`:
1. Export is working well.
2. Import says:
![image](https://user-images.githubusercontent.com/17431807/39121779-4d81d426-46fb-11e8-99cf-85b0d36c8b54.png)
I think we need that `--skipschema` because without it I'm getting:
![image](https://user-images.githubusercontent.com/17431807/39121899-c36c4ed2-46fb-11e8-9745-f1160e23f2bd.png)
So seems like we could omit activity with fixing Export/Import in Mapped BlendingEditor for a while?

warpech:
@miyconst wrote:

[How to reload with `--skip-schema` and `--mapping-mode`](https://github.com/Starcounter/StarDump/issues/118).

PatrykSzwer:
[...]

PR #212, opened 2018-05-30
Json based migration

PatrykSzwer:
Related to #181

(it's finally here! 🎆 )

First of all I would like to thank @miyconst for all help (and patient) that he provided during the process.

Those changes replace old stardump based migration mechanism with json file and wrapping classes based migration. I'm aware that this is not 100% finished version, I'll still add some changes during the review process that are described in TODO comments. Although if you can please start review as fast as you can to speed up releasing process.

I asked more reviewers for the review than usual because those changes are pretty complex and I would like to have as much different opinions and eyes that looked on that as possible.

After this is done, next step will be to merge this feature to mapping branch. It will need some changes, mostly in terms of handling keys, references to old objects etc. but it should be fairly doable to migrate this.

I also used for the first time Starcounter.Linq in this PR and I think we should start using it everywhere in this app already (After merging this PR, to avoid conflicts). This will be the first step to https://github.com/Starcounter/Blending/issues/204


src/BlendingEditor/Migration/JsonMigrationContainer.cs
```diff
@@ -0,0 +1,203 @@
...
+        public List<WebTemplateGroupMigration> Groups { get; set; }
+        public List<WebTemplateMigration> Layouts { get; set; }
+        public List<WebSectionMigration> Sections { get; set; }
+        public List<WebUrlMigration> WebUrls { get; set; }
+        public List<WebMapMigration> Maps { get; set; }
+        public List<WebUrlPropertyMigration> WebUrlProperties { get; set; }
+        public List<WebHttpHeaderMigration> WebHttpHeaders { get; set; }
+        public List<HtmlViewMigration> HtmlViews { get; set; }
+        public List<HtmlViewCompositionMigration> Compositions { get; set; }
+        public List<HtmlViewAttachmentMigration> Attachments { get; set; }
```
warpech: The property names are very inconsistent, even though your class names are consistent.

Why don't you call it:

```cs
public List<WebTemplateGroupMigration> WebTemplateGroupMigrations { get; set; }
public List<WebTemplateMigration> WebTemplateMigrations { get; set; }
....
```

or at least

```cs
public List<WebTemplateGroupMigration> WebTemplateGroups { get; set; }
public List<WebTemplateMigration> WebTemplates { get; set; }
....
```

src/BlendingEditor/Migration/MigrationClasses/HtmlViewCompositionMigration.cs
```diff
@@ -0,0 +1,52 @@
...
+        public string Html { get; set; }
+        public DateTime Created { get; set; }
+        public ulong ImportMainViewObjectNo { get; set; }
+        public ulong? NextVersionObjectNo { get; set; }
+
+        public static Dictionary<ulong?, HtmlViewComposition> ExportReferencesDictionary = new Dictionary<ulong?, HtmlViewComposition>();
+
+        public void Populate(HtmlViewComposition composition)
+        {
+            this.ImportObjectNo = composition.GetObjectNo();
```
warpech: I don't see a good pattern in naming of the properties that carry the object identity. In some migration classes it is `ImportObjectNo`, in other classes it is `ImportKey` or `<PropertyName>ObjectNo` (e.g. `AttachedViewObjectNo`), .

By choosing a single way to name this, you will make it apparent that this is a pattern used in all classes.

miyconst:
@PatrykSzwer, @warpech, I am not sure whether it worth the effort, but I could not stand the copy-pasting in this pull request.

I have prepared a solution within a single file (technically three files), which can handle any database structure as long as the database schema is a one to one match between import & export.

[miyconst/Playground@json-import-export](https://github.com/miyconst/Playground/blob/json-import-export/Playground/ImportExport/Container.cs).

Some notes:

- The code was tested with Starcounter 2.3.2 release, but it should also work with Starcounter 2.4.
- It can be enhanced to handle schema mismatches in desired way.
- It was not tested with nullable data types, such as `int?`, `long?`, but whatever obstacle happens here, it's possible to overcome.
- There are a few places which can be improved for better performance, but for such amounts of data we have in blending, it's not important.

src/BlendingEditor/BlendingEditor.csproj
```diff
@@ -273,6 +287,11 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <Import Project="$(StarcounterBin)\Starcounter.MsBuild.targets" />
+  <ProjectExtensions>
+    <VisualStudio>
+      <UserProperties Migration_4BlendingMigrationModel_1json__JSONSchema="MigrationModel.json" />
```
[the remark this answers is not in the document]
PatrykSzwer: Leftover, don't know why It's still here, since I removed file and its dependencies. I'll check that.

src/BlendingEditor/Migration/MigrationClasses/HtmlViewCompositionMigration.cs
PatrykSzwer: ImportKey - is the key of the current object that was assigned during import, so the "old" one.
PropertyNameKey - Is the "old" key of the parent for this current object.
In case of compositions objects I had to use ObjectNo's instead of keys, because there are no key property in composition objects.

PatrykSzwer: Maybe I could do some renaming to make it more understandable, like `ImportedObjectKey` for the current object keys, (ObjectNo for composition ones) and `ImportedParent{ParentClassName}ObjectKey` ?

warpech:
@miyconst thanks! My eyes were desperately looking for more DRY-ness, but I wanted to leave the pleasure of pointing that out to you :)

miyconst:
@warpech if you believe it worth the effort, I can spend some time making my solution work for Blending (if it does not work out of the box).

src/BlendingEditor/Migration/MigrationClasses/HtmlViewCompositionMigration.cs
warpech: For the current object, I would just call it `ObjectNo`. I don't see a reason for the `Import` prefix, especially since the file is an "export" file :P

For relation references I would just do `MainView = mainView.GetObjectNo()`.

I can see that @miyconst did the same here: https://github.com/miyconst/Playground/blob/json-import-export/Playground/ImportExport/Container.cs#L100

warpech:
@miyconst your generalised version looks like a really good candidate for a NuGet package. Maybe @PatrykSzwer can take it over, now that he understands the gist of it.

miyconst:
>@miyconst your generalised version looks like a really good candidate for a NuGet package.

That's true. It can be used to import/export basically anything, but before we go that way, we should make it "safe", and by that I mean we should cover quite a few corner cases which would crash with an exception unless treated properly.

PatrykSzwer:
As we spoke on the meeting with @warpech we generally like ur idea (idea with nuget package is even better) but since were a bit of a hurry, I would like to treat your solution as an improvement, create separate issue about that and do it later after this feature is merged.

PR #233, opened 2018-06-04
Merge master into mapping

kegor:
**Problem:**
After releasing Blending 6.3.0 new features should be merged into `mapping` branch.

**Solution:**
1. Firstly into `mapping` were merged changes from `master` with Blending 6.3.0 release.
2. At second into `mapping` were merged changes with `JSON-migrator`.

**Tests:**
1. Were checked manually features from the Blending 6.3.0. release description.
2. Was checked manually `JSON-migrator` feature using following steps:
- Run Blending, Reset to defaults;
- Create a new group, move TopBar layout into a new group, make export file.
- Remove newly created group, delete TopBar layout, Delete all compositions.
- Import JSON dump file.
Result:
- Group and TopBar layout were recreated.
- ⚠️ for the moment TopBar layout was not visible, working on this.


src/BlendingEditor/Migration/MigrationClasses/WebHttpHeaderMigration.cs
```diff
@@ -0,0 +1,32 @@
...
+        public string Name { get; set; }
+        public string Value { get; set; }
+        public bool PartialMatching { get; set; }
+        public string ImportWebUrlKey { get; set; }
+
+        public void Populate(WebHttpHeader webHttpHeader)
+        {
+            this.Name = webHttpHeader.Name;
+            this.Value = webHttpHeader.Value;
+            //this.PartialMatching = webHttpHeader.PartialMatching;
```
PatrykSzwer: Why WebHttpHeader class doesn't have PartialMatching field? How this functionality is handled without it? It should be there.

src/BlendingEditor/Migration/MigrationClasses/WebUrlPropertyMigration.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public void Populate(WebUrlProperty webUrlProperty)
+        {
+            //this.Name = webUrlProperty.Name;
+            this.ImportWebUrlKey = webUrlProperty.Url.Guid;
+        }
+
+        public void RecreateDbEntity(Dictionary<string, WebUrl> webUrlReferencesDictionary)
+        {
+            new WebUrlProperty
```
PatrykSzwer: This also should have check for already existing same entity just like all other similar methods in all migration classes. To achieve that it will also need guid to be migrated.

src/BlendingEditor/Migration/MigrationClasses/WebHttpHeaderMigration.cs
```diff
@@ -0,0 +1,32 @@
...
+            this.ImportWebUrlKey = webHttpHeader.Url.Guid;
+        }
+
+        public void RecreateDbEntity(Dictionary<string, WebUrl> webUrlReferencesDictionary)
+        {
+            new WebHttpHeader
```
PatrykSzwer: This also should have check for already existing same entity just like all other similar methods in all migration classes. To achieve that it will also need guid to be migrated.

src/BlendingEditor/Migration/MigrationClasses/WebHttpHeaderMigration.cs
kegor: I added it now. I think it was not here because WebHttpHeader class was not synced with the master branch from the time of its creation by Habib.

kegor:
Ready for review.

PatrykSzwer:
Currently I can't build `Blending.Mapper` due to error:
```
[MGEN] DEBUG    Compiling mapper ...
C:\repositories\Mapping\Blending.Mapper\cpp\WebUrlProperty.cpp:736:50: error: no member named 'on_update_Guid' in
      'mapper<BlendingProvider::DataModels::WebMap>'
                        mapper<BlendingProvider::DataModels::WebMap>::on_update_Guid(root, value);
                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
1 error generated.
[MGEN] ERROR    Mapper compilation failed.
```

warpech:
> Maybe it will be better to make a separate issue for this issue because it is not strongly related to the moving on a mapping branch?

Yes, please

PatrykSzwer:
Although without this, import/export won't work correctly due to unconnected websections to webtemplates.

Issue #69, opened 2018-02-07
Implement DB Mapping, make the app usable for the Apps Team


LarsTjernstrom:
# What to do
- [x] Replace Simplified with Local db Model (https://github.com/Starcounter/RebelsLounge/issues/240)
- [x] Refactor so the Authentication works with on branch without Simplified. See any of the apps, People, Search, UserAdmin with the branch without Simplified to see how it's done right now
- [x] Design UI Wireframe complete (https://drive.google.com/drive/u/0/folders/12TeSCAPs0v5unQ-CnckDCqelpTCj5qP4)
- [x] Implement "Underware" (https://github.com/Starcounter/Blending/issues/50)
- [x] Pixel perfect UI design complete (https://github.com/Starcounter/Blending/issues/50)
- [x] "Feature complete" (?)
- [ ] Tests pass
- [ ] Released to Appstore

##### Browsers affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari

warpech:
Thank you @LarsTjernstrom for expressing your expectations towards the next release. For the issues that are already tracked, I have added a reference link.

Some items that I marked as `{?}` require clarification from you:

- what are the problems with Authentication? it works. If you mean that some changes are needed to make it work with mapping, please provide more information
- what do you mean by the point "Feature complete"? Blending app suite has the whole roadmap ahead: https://github.com/Starcounter/CompanyTrack/blob/master/WebPlatform/roadmap-2018q1.md. Do you mean: make sure all features work in the mapping branch?

LarsTjernstrom:
Refactor so the Authentication works with a mapped version of SignIn. If it's allready complete, just check the checkbox
Feature complete is not my input, it's Stefans requirement...but to be complete, you have to have a set of features specified to be able to complete them (and I don't think we have).

warpech:
I will then just assume that it means current features should work on the branch without Simplified.

Please note that any progress in this branch is blocked until https://github.com/Starcounter/RebelsLounge/issues/240 is done.

warpech:
Marked `Implement "Underware" (#50)` as done

warpech:
The remaining tasks are:

- Tests do not pass (https://github.com/Starcounter/Blending/issues/182)
- Import feature does not work (https://github.com/Starcounter/Blending/issues/181)

After the above issues done, we can release a version to the App Warehouse and verify with the apps team that all is done. @malx122 will like to be updated about this, too.

[...]

PR #248, opened 2018-06-18
Problem: Blending tests in the mapping branch should pass.

kegor:
**Problem:**
This PR contains changes for fixing the Blending tests in the `mapping` branch, originally from https://github.com/Starcounter/Blending/issues/69#issue-295083517.

**Solution:** 
- Empty blend.json files were added into the pair of test helper apps.
- Select query in DataHelper was rewritten.
- References to the `Simplified` were removed.

**Tests:**
Passed: https://teamcity.starcounter.org/viewLog.html?buildId=58286&tab=buildResultsDiv&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24.


PatrykSzwer:
BTW could you also remove commented code from here before you merge? https://github.com/Starcounter/Blending/blob/mapping/src/BlendingEditor/Website/Helpers/MigrationHelper.cs#L67
I forgot to delete it when I was changing stardump based migration to the json based one.

kegor:
Hi @PatrykSzwer, I definitely need to explain this trick with the empty blend.json files:
I found that the problem like this `System.ArgumentException: The following URI is already added to blender: "/blendingeditor/app-name" on token "JustEmptyToken"` here https://teamcity.starcounter.org/viewLog.html?buildId=58153&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24&tab=buildLog

I think because `BlendingEditor_TestHelper` copying the `blend.json` from the `BlendingEditor` debug folder because it has the reference on the BE project and because `blend.json` has property copy local.
The problem here is that when `BlenderEditor_TestHelper` starts the `Blender` class trying to create attaching rules from the copied `blend.json` file again.

By adding empty `blend.json` files into the pair of the test projects I overrided this behaviour.

PatrykSzwer:
I think it should be either reported or documented somewhere, but I'm not sure where. @warpech do u have some opinion about it?

warpech:
Is it required for `blend.json` to have property copy local? Maybe it has to be fixed in our `csproj`, not documented? If you think it deserves further discussion, pls make an issue

Issue #241, opened 2018-06-12
Add GUI to activate/deactivate of default rules for apps


warpech:
The GUI will present the default rules in form of a table. The only editable part is a checkbox that turns the rule to be active or not. 

The table should distinguish from the initial state and the current state. The initial state is the one that is registered through `blend.json` or `Blender` API. 

As part of this task, a new data model class needs to be created. This class will hold the data needed to identify default rules registered through `blend.json` or `Blender` API.

This configuration should be part of import/export.

This work should only be done in the `mapping` branch.

UI will look somehow like:

```
Uri                            | AppName    | Contexts             | Dir   | Active
/people/partial/settings/{?}   | People     | fullpage, settings   | -->   | [ ]   
```

kegor:
According to this https://github.com/Starcounter/Blending/pull/250#issuecomment-404127350 additional changes will be tracked in a new PR.

kegor:
Closing because additional changes were merged in the #259.

PR #250, opened 2018-06-27
Problem 241: attachment rules page with possibility to only activate/deactivate.

kegor:
**Problem #241:**
For the start, we would like to manage the state of the attachment rules right from the BlendingEditor page.

**Solution:**
1. `AttachmentRule` entity was added into both BE/BP data models namespace. Objects are creating in the DB on the BE app start for now.
2. `AttachmentRules` page was added, it shows almost all of `AttachmentRule` entity fields.
3. After clicking on the `Active` check-box you are changing state of the rule both in DB and in the Blender class internal dictionary.
4. Attachment rules which were exported into JSON-file can be successfully imported back.

**Screenshots:**
image

**Notes:**
I pushed needed changes in the `Blending.Mapper` and `DoctorM` repositories.


src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -84,6 +85,63 @@ public static void GenerateData()
...
+        }
+
+        public static void CreateAttachingRule(BlendingInfo biValue)
+        {
+            var attachingRule = new AttachingRule
+            {
+                Guid = Guid.NewGuid().ToString(),
+                Uri = biValue.Uri,
+                AppName = biValue.AppName,
+                Direction = GetDirectionCode(biValue),
```
warpech: I dislike this, because it is denormalized. I think we should have two boolean properties of the same name that is used in Blender API: `AllowFromDirection: true|false`, `AllowToDirection: true|false`

Rendering it as `<-->`, `<--`, etc, should be performed on the client.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -84,6 +85,63 @@ public static void GenerateData()
...
+                Contexts = biValue.Contexts.Length > 0 ? string.Join(", ", biValue.Contexts) : "",
```
warpech:
```
Contexts = string.Join(", ", biValue.Contexts),
```

would have exactly the same result

src/BlendingProvider/DataModels/AttachingRule.cs
```diff
@@ -0,0 +1,20 @@
...
+        /// Instead of deleting items, this boolean should be toggled to false.
+        /// </summary>
+        public string Guid { get; set; }
+        public string AppName { get; set; }
+        public string Direction { get; set; }
+        public string Uri { get; set; }
+        public bool IsActive { get; set; }
+        public string Contexts { get; set; }
+        public string Token { get; set; }
+    }
```
warpech: We should not have the two classes `AttachingRule` and `WebMap`, because they are essentially the same.

It is ok to have a new class, but then `WebMap` should be removed and area items functionality should be changed to use the new class.

src/BlendingEditor/wwwroot/BlendingEditor/views/MasterPage.html
```diff
@@ -9,6 +9,7 @@
             <a slot="blendingeditor/menu-item" href="/BlendingEditor/layoutconditions">Conditions</a>
             <a slot="blendingeditor/menu-item" href="/BlendingEditor/collections">CSS Variables</a>
             <a slot="blendingeditor/menu-item" href="/BlendingEditor/compositioneditor">Composition Editor</a>
+            <a slot="blendingeditor/menu-item" href="/BlendingEditor/attachingrules">Attaching rules</a>
```
warpech: According to the docs, these feature is called "attachment rules", not "attaching rules".

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachingRulesPage.html
```diff
@@ -0,0 +1,68 @@
+<link rel="import" href="/sys/juicy-select/juicy-select.html">
+<link rel="preload" as="style" href="/sys/uniform.css/uniform.css">
+<link rel="import" href="/sys/vaadin-combo-box/vaadin-combo-box.html">
```
warpech: `juicy-select` and `vaadin-combo-box` are imported but not used

src/BlendingEditor/wwwroot/BlendingEditor/views/MasterPage.html
kegor: Fixed.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachingRulesPage.html
kegor: Fixed.

src/BlendingProvider/DataModels/AttachingRule.cs
kegor: We decided to make migration from `WebMap` to the `AttachmentRule` in a separate PR.

src/BlendingEditor/Website/Helpers/DataHelper.cs
kegor: I saw exception in some particular cases which said about null value in the `string.Join(", ", biValue.Contexts)` this why I added this check, but I agree that it will be better to find the problem in this case rather than just mute it.

src/BlendingEditor/DataModels/AttachingRule.cs
```diff
@@ -0,0 +1,19 @@
+﻿using Starcounter;
+
+namespace BlendingEditor.DataModels
+{
+    [Database]
+    public class AttachingRule
+    {
+        /// <summary>
+        /// Instead of deleting items, this boolean should be toggled to false.
```
PatrykSzwer: This comment seems misplaced, since it says about boolean but it is above string (Guid).

src/BlendingEditor/Migration/JsonMigrationContainer.cs
```diff
@@ -23,6 +23,7 @@ public class JsonMigrationContainer
         public List<CssVariableMigration> CssVariables { get; set; }
         public List<CssVariableValueMigration> CssVariableValues { get; set; }
         public List<CssCollectionRelationMigration> CssCollectionRelations { get; set; }
+        public List<AttachingRuleMigration> AttachingRules { get; set; }
```
PatrykSzwer: Regarding @warpech comment, should this be renamed to `Attachment`, not `Attaching`?

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -84,6 +85,63 @@ public static void GenerateData()
...
         }
 
+        public static void PopulateAttachingRules()
+        {
+            var blendingInfosDictionary = Blender.ListAllByUris();
+
+            Db.Transact(() =>
+            {
+                // TODO: rewrite to LINQ
+                var attachingRules = Db.SQL<AttachingRule>($"SELECT ar FROM {typeof(AttachingRule)} ar");
```
PatrykSzwer: To represent this as a LINQ all u need is `DbLinq.Objects<AttachingRule>()`

src/BlendingEditor/DataModels/AttachingRule.cs
kegor: Yeah, exactly, I removed it already!

src/BlendingEditor/Migration/JsonMigrationContainer.cs
kegor: Yes, this was renamed.

src/BlendingEditor/Website/Helpers/DataHelper.cs
kegor: Fixed, thank you for reminding.

src/BlendingEditor/Website/Helpers/DataHelper.cs
kegor: Thank you, Tomek, I will try!

tomalec: Then you can add it as other elements to https://github.com/Starcounter/Blending/tree/master/src/BlendingEditor/wwwroot/BlendingEditor/elements

kegor: Fixed comment from Marcin, but seems like I need to think a little about the adding it into `elements` folder, maybe you could first take a look on my using of your suggestion (when I'll make a push)?

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -15,51 +15,101 @@ <h2 slot="blendingeditor/attachingrules-heading">Attaching rules</h2>
...
+                                <template is="dom-if" if="{{!attachmentRule.AllowFromDirection}}">
+                                    <template is="dom-if" if="{{attachmentRule.AllowToDirection}}">
+                                        <blending-direction-arrow from></blending-direction-arrow>
+                                    </template>
+                                </template>
+                                <template is="dom-if" if="{{attachmentRule.AllowFromDirection}}">
+                                    <template is="dom-if" if="{{attachmentRule.AllowToDirection}}">
+                                        <blending-direction-arrow from to></blending-direction-arrow>
+                                    </template>
+                                </template>
```
tomalec: Instead of lines 30-44, you can write just

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -15,51 +15,101 @@ <h2 slot="blendingeditor/attachingrules-heading">Attaching rules</h2>
...
+    <script>
+        (function () {
+            customElements.define('blending-direction-arrow', class extends HTMLElement {
+                constructor() {
+                    super();
+                    this.attachShadow({ mode: 'open' }).innerHTML = `
+                    <style>
+                    :host{
+                    display: inline-block;
+                    }
```
tomalec: Could you indent those lines to make it more readable?

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -15,51 +15,101 @@ <h2 slot="blendingeditor/attachingrules-heading">Attaching rules</h2>
...
                 Save
             </button>
-            <button slot="blendingeditor/attachingrules-cancel-button" value="{{model.CancelChanges$::click}}" onmousedown="++this.value;" disabled="{{!model.Trn.IsDirty}}" title="Cancel all changes">
+            <button slot="blendingeditor/attachmentrules-cancel-button" value="{{model.CancelChanges$::click}}" onmousedown="++this.value;" disabled="{{!model.Trn.IsDirty}}" title="Cancel all changes">
                 Cancel
             </button>
         </template>
     </dom-bind>
+    <script>
+        (function () {
```
tomalec: Technically this IIFE is not needed here

src/BlendingEditor/Website/Helpers/DataHelper.cs
tomalec: YW, done.

src/BlendingEditor/Website/Helpers/DataHelper.cs
kegor: A redundant comment was removed locally as well.

kegor:
@warpech, @PatrykSzwer, @tomalec I have updated the PR description. I have added notes about the manual check, feel free to correct me, if you'll find some steps redundant or invalid.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done, thank you.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Tomek, maybe strange but with this simplification column is fully empty. I checked this manually. And after reverting to the old huge code all works again. Maybe this is because `customElement` looking only on single `from`, `to` and not on `from="true"` or `to="false"`?

tomalec: My bad it should be `$=`:
```html
<blending-direction-arrow from$="{{attachmentRule.AllowFromDirection}}" to$="{{attachmentRule.AllowToDirection}}"></blending-direction-arrow>
```

PatrykSzwer:
There are some conflicts in the PR, please fix.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -0,0 +1,113 @@
...
+                                    </template>
+                                </template>
+                                <template is="dom-if" if="{{!attachmentRule.AllowFromDirection}}">
+                                    <template is="dom-if" if="{{attachmentRule.AllowToDirection}}">
+                                        <blending-direction-arrow from></blending-direction-arrow>
+                                    </template>
+                                </template>
+                                <template is="dom-if" if="{{attachmentRule.AllowFromDirection}}">
+                                    <template is="dom-if" if="{{attachmentRule.AllowToDirection}}">
+                                        <blending-direction-arrow from to></blending-direction-arrow>
```
tomalec: Are you sure this does not work:
```html
<blending-direction-arrow from$="{{attachmentRule.AllowFromDirection}}" to$="{{attachmentRule.AllowToDirection}}"></blending-direction-arrow>
```
https://steady-look.glitch.me/
https://glitch.com/edit/#!/steady-look?path=index.html:46:54

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -0,0 +1,99 @@
...
+                        }
+                    </style>
+                    <div class="to"><--</div>
+                    <div class="from">--></div>
+                    <div class="none">--</div>
+                    <div class="both"><--></div>
+                `;
+            }
+        });
+    </script>
```
tomalec: I'd put custom element definition either to a separate file, or to the top of this file, so it would be defined before used in dom-bind.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Hi @tomalec, perhaps you somehow looked into the previous iteration of the code, in the current one I already left your one-line solution https://github.com/Starcounter/Blending/pull/250/commits/859e885927e06904abf8038e595f496b304aa9ea#diff-326a39b09a0cf214bafe1fdbf1c28255R30, it works as expected locally.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done.

kegor:
I'm approving both @PatrykSzwer's findings:
1. I also see similar `dancing` for color-picker, so this maybe the common problem in the markup, maybe @tomalec or @alshakero could point the possible reason more quickly than me? If no I will investigate this.
dancing-color-picker

2. In my understanding this issue mostly related to the next step `moving from WebMaps to AttachmentRules` (but I agree that in current PR `AttachmentRules` looks a bit useless), anyway I'll try to describe my understanding the Patryk's case here and maybe we'll find the way:
2.1. Commonly if you'll start `SignIn.Mapper` first and then `Blending.Mapper` then you'll see this row in the `AttachmentRules` table:
image
2.2. `BlendingEditor` took this line on start from `Blender` class, and `Blender` has it because of this line: https://github.com/Starcounter/SignIn/blob/a3a5ecc7d3a1af4ad3ecb99b179233171b194dee/src/SignIn/Api/BlenderMapping.cs#L10
2.3. If you'll made action like `Reset to defaults` which will create new `Layouts/Sections/WebMaps/e.t.c` then following code will be called: https://github.com/Starcounter/Blending/blob/2865a6f24cb995ffe293c0bca167d300d8035028/src/BlendingProvider/Website/Api/BlendingHandlers.cs#L298 which will create several new blending rules (in `Blender` class as well though the `Blender.MapUri() function) which will be related to both `Layouts,  sections` e.t.c and to the `foreignUris` like `/signin/user` but we will not see them, becuase BE refreshes `AttachmentRules` in current version only on start.
2.4. And when Patyk tried to disable `/signin/user', he was disabling the `AttachmentRule` which was not in use, on my view, and because real `AttachmentRule` which was in use (with more complex `Token`-string btw) was not in the table at all.

I need to think about the conclusion a little. To be continued.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -0,0 +1,100 @@
...
+                    <div class="both"><--></div>
+                `;
+        }
+    });
+</script>﻿
+
+<template>
+    <dom-bind>
+        <template is="dom-bind">
+            <h2 slot="blendingeditor/attachmentrules-heading">Attaching rules</h2>
```
warpech: The text in H2 should be "Attachment rules"

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -0,0 +1,100 @@
...
+                            display: inline-block;
+                        }
+                        :host(:not([to]):not[from]) .none {
+                            display: inline-block;
+                        }
+                    </style>
+                    <div class="to"><--</div>
+                    <div class="from">--></div>
+                    <div class="none">--</div>
+                    <div class="both"><--></div>
```
warpech: It is wrong to use `<` and `>` in text in HTML.

These characters should be either escaped (`&lt;` and `&gt;`), or even better, replaced with nice Unicode arrows.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done.

kegor:
@warpech could you please clarify your case with restarting the BlendingEditor?
I tried to do:
1. Stop BlendingEditor on the page `http://127.0.0.1:8181/#/databases/default`.
2. Start BlendingEditor on the page `http://127.0.0.1:8181/#/databases/default/applicationStart` specifying the BlendingEditor.exe path and path to the `wwwroot` folder, both in Blending sub-module:

But I'm getting this error:
image

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
```diff
@@ -25,10 +26,10 @@
...
                         }
                     </style>
-                    <div class="to"><--</div>
-                    <div class="from">--></div>
-                    <div class="none">--</div>
-                    <div class="both"><--></div>
+                    <div class ="to">←</div>
+                    <div class ="from">→</div>
+                    <div class ="none">–</div>
+                    <div class ="both">↔</div>
```
warpech: Should be `class=`, not `class =`

warpech:
I just restarted it by calling `star src/blendingeditor/bin/debug/blendingeditor.exe` in the command line.

src/BlendingEditor/wwwroot/BlendingEditor/views/AttachmentRulesPage.html
kegor: Done.

kegor:
About the all active rules after restarting the `BlendingEditor`:
I think that the answer is related to this line:  https://github.com/Starcounter/Blending/pull/250/files#diff-299d9aaf56f8ace9689c69d97a1b7561R97
But for now I can't answer why previously deactivated in the Blender class inner dictionary rules becomes Active again after invoking `ListAllByUris()` (I plan to discuss this with Alexey).

Also I met this problem during testing this case:
image
And I finally decided return this check: https://github.com/Starcounter/Blending/pull/250/files#diff-299d9aaf56f8ace9689c69d97a1b7561R118.

For the rules with null Contexts like this:
image

warpech:
@kegor I think that this can be merged as it is. Please make an issue that contains the remaining problems and we will track it in a new PR.

alemoi:
> The inactive rules are activated again after BlendingEditor restarts.

If `BlendingEditor` not storing the changed settings and re-applying them in `Main` I guess the rules will not be changed.

Issue #239, opened 2018-06-12
Rename "CSS Variables" to "Theme configurations"


warpech:
As agreed on the meeting today, rename "CSS Variables" to "Theme configurations".

This includes the UI and the names of the files as well as the explanation in the [docs](https://github.com/Starcounter/Blending/blob/master/docs/gui.md)

warpech:
Closing as done

PR #251, opened 2018-06-27
Rename CSS Variables, CSS Collections to CSS custom properties and Theme configurations

PatrykSzwer:
For #239


PatrykSzwer:
I talked with @warpech and we agreed to **not** change names of database entities, due to problems that would be raised (changing mappers, backward compatibility etc.).

Issue #203, opened 2018-05-23
REST interface for importing of settings


warpech:
After #181 is done, add a REST `PUT` handler that replaces the current settings with the configuration sent as a JSON payload.

@malx122 reported today on the "App developer experience ref group" meeting that he needs a programmatic way to import the settings.

warpech:
@malx122 says he will need this fast, because he resets the db 20 times a day. We should do it in June.

kegor:
If we are talking about the ability to update the settings using tools such as [Postman](https://www.getpostman.com/), then the solution seems straightforward to me: register the `PUT` handler as an example here: https://github.com/Starcounter/level1/blob/09ebe2cd524f08034acd469059b99804124a648f/src/Starcounter.Administrator/Server/API/RestHandlers.cs#L293 in which the JSON-line will be used to update the settings in the database.

Another question for me here appeared in the fact that I do not see much point in updating the `WebSettings` of the `BlendingEditor`, since they contain only one significant field for turning off the `ViewKeeper`, maybe we are more likely speaking about possible updating settings in any application?

@warpech could you please clarify the final target a little?

warpech:
The user story is that a solution owner would like to install apps on a fresh installation of Starcounter. After the apps are installed, he wants to import Blending app suite settings in a script. Most likely we will push a JSON file to a HTTP endpoint using a tool like cURL.

> Another question for me here appeared in the fact that I do not see much point in updating the WebSettings of the BlendingEditor, since they contain only one significant field for turning off the ViewKeeper, maybe we are more likely speaking about possible updating settings in any application?

We are talking about importing the file that is the result of a manual export in BlendingEditor, with anything that can be expected there.

kegor:
@warpech I've pushed the few lines of code to finish clarifying the final target here: https://github.com/Starcounter/Blending/commit/40f774752a381f7c4c1cc49c30a1d8255120c54c.
And this works like this:
![import-using-put2](https://user-images.githubusercontent.com/17431807/41879102-ba45ed24-78e0-11e8-884f-91abc55b68ca.gif)
Is this looks like right direction?
(I used Postman bc I'm not familiar with the cURL as the script execution tool)

warpech:
Yes, that's it!

kegor:
Closing because related changes were merged in the #252.

PR #252, opened 2018-06-27
Problem 203: REST interface for importing of settings.

kegor:
**Problem #203:**
>@malx122 reported today on the "App developer experience ref group" meeting that he needs a programmatic way to import the settings.

**Solution:**
1. PUT handler was added.
2. `ImportDataToDabatase` method was rewritten for the possibility to use passed JSON-string in the import process.

***Notes:***
1. If I'm using `Postman` for sending PUT request then I can see the blocking message about already existing Compositions, but I'm not sure about other tools like `cURL`.
2. If I'm using case in which I firstly make JSON-dump file on the default data and then delete some layouts, then after successful import I will not see them.
2.1. From the one view, this is according to the initial migration logic, because deleted Layouts are still in the DB, but with `Deleted = true`.
2.2. From the other side, this behavior might surprise some of the users.

**Screenshots:**
_This GIF was made for the previous version with programmatic clearing data, but for now I decided remove this, because this contradicts initial migration logic implemented by Patryk:_
import-using-put2

_If I have Compositions in both passed JSON-string and in the DB then I still can see the error message about it:_
image

**Tests:**
✅ A manual test was successful.
⚠️ Unit-tests are not presented yet.


src/BlendingEditor/Website/Api/MainHandlers.cs
```diff
@@ -37,6 +37,13 @@ public void Register()
                 return "All data of layouts were deleted.";
             });
 
+            Handle.PUT("/blendingeditor/updatedata", (Request request) =>
```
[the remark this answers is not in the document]
kegor: Yeah, exactly, I just wanted to discuss this, and I agree that first let the `SignIn` action be produced.

src/BlendingEditor/Website/Api/MainHandlers.cs
```diff
@@ -37,6 +37,13 @@ public void Register()
...
+            {
+                string message;
+                MigrationHelper.ImportDataToDabatase(null, request.Body, out message);
```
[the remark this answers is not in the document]
kegor: The new manual check with using cookies:
import-using-put_with_cookie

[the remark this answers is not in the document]
kegor: @warpech I've added few lines which are checking hardcoded bearer token in the request headers, this is working like this:
import-using-put_with_bearer_token
Could you please clarify:
1. Is this way more scriptable for David? (I tried to ask him about additional endpoint for the export but didn't answer, I guess he is very busy).
2. Is this way in the same direction which you are mentioned above?
3. WDYT about using third-party token-auth packages for realizing such things for this task?

warpech:
I hope that on Tuesday we will be able to present to @malx122 that this is done and already available in `mapping`.

src/BlendingEditor/Website/Helpers/ArgsHelper.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
+
+namespace BlendingEditor.Website.Helpers
+{
+    public class ArgsHelper
+    {
+        private static string _token = "mysecrettoken";
```
kegor: @warpech I'm not sure about the best case here in the current implementation, maybe I should make it `String.Empty` and check that it is not empty while checking token, so this `_token` becomes required for the import through PUT request process?

kegor: Also about the scripting:
I just added `token=123` for my manual test in the default `run.bat` file:
`star %* --resourcedir="%~dp0src\BlendingEditor\wwwroot" "%~dp0src\BlendingEditor\bin\%CONFIGURATION%\BlendingEditor.exe" token=123`

Because fairly speaking I didn't understand the usage purpose for the rest of the parts in your example `\-api\-token=mysecrettoken`, could you please clarify your idea here a little?

kegor: And seems like the tests are fallen because TC tried to build current changes on the level1 version with this commit included https://github.com/Starcounter/level1/commit/407530749dca4628be954a3d626c23e9d50d257c. 

I tried to add using of `Starcounter.Advanced` into `AreasPage.json.cs` with installed `#2.4.990.117 (07 Jul 18 01:05)` from here https://teamcity.starcounter.org/viewLog.html?buildId=60185&tab=artifacts&buildTypeId=StarcounterDevelop_DevelopNightlyWindowsMapping and it looks buildable again. Should we add this new namespace in the all related files in our both Blending apps?

kegor:
> Can you give some example? Your commit 7148b13 looks fine without any 3rd party packages.

Nice to hear :) actually I meant something from here https://www.nuget.org/packages?q=token+auth or something built-in in ASP.NET Core like https://www.nuget.org/packages/Microsoft.AspNetCore.Authentication.JwtBearer/ but for now I think that this looks too much for the current task.

src/BlendingEditor/Website/Helpers/ArgsHelper.cs
PatrykSzwer: Yes, please add this namespace if its needed for newest SC version.

warpech: > Should we add this new namespace in the all related files in our both Blending apps?

@kegor you were negotiating the new API with @alemoi so I think there is no need to double-ask how to make it work with our apps.

warpech: My `\-api\-token=mysecrettoken` has exactly the same purpose as your `token=123`. In my case I escaped the `-` characters, because, according to the docs document that I linked, they need to be escaped.

warpech: The app should be secure by default, so I think this means the default value of the token must be empty and thus ignored.

See:
- https://github.com/Starcounter/CompanyTrack/issues/286

kegor: @warpech I sent changes related to the empty default value of the token.

warpech:
Please add usage instructions to `docs/administration.md` and I think it can be merged then.

docs/administration.md
```diff
@@ -10,3 +10,7 @@ Calling this URI replaces all the current app data (layouts, layout groups, area
 
 * **TopbarLayout** - a layout with two areas: `Topbar` and `Main`. The default route-all rule uses this layout.
 * **LoginLayout** - a layout with three areas: `Footer`, `Logo` and `SignIn` prepared to use with SignIn app.
+
+## [`/blendingeditor/updatedata`](http://localhost:8080/blendingeditor/updatedata)
+
+This URI can be used in a PUT-request to update the configuration data. To use this feature, the application must be started with passing the `-api-token` argument: `star.exe src/blendingeditor/bin/debug/blendingeditor.exe \-api\-token=mysecrettoken`. The request body must contain a JSON string with the new data, and the authorization header must contain a bearer token with the value used for the passed `-api-token`.
```
warpech: Can you please add an example of usage with cURL?

kegor: Done, @warpech could you please check updated text when you'll find a time, or I can start to merge right now?

warpech: Looks great. You can merge, as it is already approved.

PR #253, opened 2018-06-28
Problem 140: JSON-based migration for WebSettings. 

kegor:
**Problem #140:**
In continuation of https://github.com/Starcounter/Blending/pull/176.

**Solution:**
1. WebSettings migration class was added.
2. To avoid the creation default `WebSettings` on import, invocation of the `EnsureSettings` method was moved to the `Program.cs`, `SettingsPage` now uses only `GetSettings` method.
3. `WebSettings.Guid` field was added for a more clear check of the existing `WebSettings`.
4. Current implementation still based on the following general case: if we have same settings in the dump, then we should update existing, if no then we should leave existing as is.

If we will agree with the Guids here I also should push local related changes into `Blending.Mapper` and `DoctorM` repos.

**Screenshots:**
import-viewkeeper-settings3

**Tests:**
✅ Manual check of the general case was successful.


warpech:
@kegor is this done? A review was not requested

kegor:
Yes, sorry, this can be reviewed. Also, the important notice that this PR depends on additional changes in the `Blending.Mapper` and `DoctorM` repos (related to `WebSettings.Guid` explained in PR description).

warpech:
I would like to move fast with the changing of the branch names.

May I recommend that you merge it as it is and we review it retrospectively?

@PatrykSzwer WDYT?

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -581,14 +581,23 @@ public static WebSettings EnsureSettings()
...
+                        Description = "Settings for the BlendingEditor and BlendingProvider apps"
                     };
                 });
             }
             return settings;
         }
 
+        public static WebSettings GetSettings()
+        {
+            var settings = Db.SQL<WebSettings>($"SELECT s FROM {typeof(WebSettings).FullName} s WHERE s.{nameof(WebSettings.Deleted)} = ?", false).FirstOrDefault();
```
kegor: This need to be replaced with LINQ statement before the merge.

PR #255, opened 2018-07-10
Adding Starcounter.Advanced namespace

kegor:
**Problem:**
- We need to update using of the `Blender` class in the both Blending apps after this commit https://github.com/Starcounter/level1/commit/407530749dca4628be954a3d626c23e9d50d257c.
- Also, we need the new features from the updated `Blender` class for improving this PR https://github.com/Starcounter/Blending/pull/250.

**Solution:**
- `null` value was added in most of the invocations of the `Blender` API methods.
- `Starcounter.Advanced` namespace was added into appropriate files.
- `%` was changed on a `^` in the Blending function for constructing token-string to avoid conflicts with the `Blender` logic.


kegor:
Related issue: https://github.com/Starcounter/UniformDocs/issues/332.

Issue #209, opened 2018-05-28
Add `palindrom-error-catcher` to TopBarLayout


warpech:

warpech:
@alshakero sasy that we changed our minds about it. I will check that.

alshakero:
Ping.

warpech:
The intention of https://github.com/Palindrom/palindrom-client/issues/69 was to move the rendering of the disconnection bar from Starcounter app shell (palindrom-client) to solution layout.

So far we indeed succeeded in removing the UI from Starcoutner app shell, but now we show zero feedback in the UI. This leaves us at point 5 of this guideline: https://ux.stackexchange.com/a/81423

Adding `palindrom-error-catcher` to **all layouts** (not just TopBarLayout) is the best way to advance directly to point 2.

I don't think that is the final solution, but it is a step towards the final solution.

alshakero:
It should be noted that two instances of `palindrom-error-catcher` means the error showing twice.

alshakero:
So I don't fully understand, wouldn't it be an issue to have the error showing twice? Doesn't "Adding `palindrom-error-catcher` to **all** layouts" imply multiple instances of it in the same page?

warpech:
> Doesn't "Adding palindrom-error-catcher to all layouts" imply multiple instances of it in the same page?

In the default config, you never get more than one layout at the screen at the same time.

PR #261, opened 2018-07-19
Add palindrom-error-catcher to default layouts

alshakero:
I don't understand why but importing HTML element doesn't work so I inlined the element as is. 

Fixes: https://github.com/Starcounter/Blending/issues/209


warpech:
> I don't understand why but importing HTML element doesn't work so I inlined the element as is.

Use [`<enlighted-link>`](https://github.com/juicy/enlighted-link). Explanation: https://github.com/Starcounter/RebelsLounge/issues/263. 

There was a big fuzz about it.You were the reviewer of a PR that added it to SCF

alshakero:
>  or directly in the composition editor?

I think we can create a check in composition editor that triggers when you it detects a `<link rel="import">` when you click Save and shows an panel that looks like this:

An HTML import was detected in your composition, please use [`<enlighted-link>`](https://github.com/juicy/enlighted-link) instead of `<link />`. _[why?](https://github.com/Starcounter/RebelsLounge/issues/263])_

alshakero:
> I think it should be added to the app shell. 

It's [there ](https://github.com/Starcounter/StarcounterClientFiles/blob/3.x/src/StarcounterClientFiles/wwwroot/sys/starcounter.html#L3)already.

I pushed a new commit that uses it.

PR #266, opened 2018-07-30
Improvements to the administrator's API

warpech:
Changes are explained in the commit messages

This PR is required to automate loading of the configuration for https://github.com/Starcounter/HeadsOmni/issues/61


warpech:
@kegor some tests are failing. Is that connected with my changes?

src/BlendingEditor/Website/Helpers/AuthorizedHandle.cs
```diff
@@ -29,6 +29,14 @@ public static void PUT<T>(string putUrl, Func<T, Response> code)
             });
         }
 
+        public static void DELETE<T>(string putUrl, Func<T, Response> code)
```
kegor: [Minor:] maybe rename `putUrl` -> `deleteUrl` for consistent view?

docs/administration.md
```diff
@@ -1,17 +1,17 @@
 # Administration
 
-## [`/blendingeditor/cleardata`](http://localhost:8080/blendingeditor/cleardata)
+## [`DELETE /blendingeditor/cleardata`](http://localhost:8080/blendingeditor/cleardata)
 
 Calling this URI deletes all the current app data (layouts, layout groups, areas, area items and layout conditions).
 
-## [`/blendingeditor/resetdata`](http://localhost:8080/blendingeditor/resetdata)
+## [`PUT /blendingeditor/resetdata`](http://localhost:8080/blendingeditor/resetdata)
```
kegor: I would say that using `PUT` HTTP-verb here is not very suitable, because `PUT` mostly used when data should be updated or created by the passed data (https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol#Request_methods).

docs/administration.md
```diff
@@ -1,17 +1,17 @@
...
 
 Calling this URI replaces all the current app data (layouts, layout groups, areas, area items and layout conditions) with the defaults, which are:
 
 * **TopbarLayout** - a layout with two areas: `Topbar` and `Main`. The default route-all rule uses this layout.
 * **LoginLayout** - a layout with three areas: `Footer`, `Logo` and `SignIn` prepared to use with SignIn app.
 
-## [`/blendingeditor/updatedata`](http://localhost:8080/blendingeditor/updatedata)
+## [`PUT /blendingeditor/updatedata`](http://localhost:8080/blendingeditor/updatedata)
 
 This URI can be used in a PUT-request to update the configuration data. To use this feature, the application must be started with passing the `-api-token` argument: `star.exe src/blendingeditor/bin/debug/blendingeditor.exe \-api\-token=mysecrettoken`. The request body must contain a JSON string with the new data, and the authorization header must contain a bearer token with the value used for the passed `-api-token`.
```
kegor: This description is now true for all mentioned actions above, so maybe it will be useful mention this details below the each?

docs/administration.md
warpech: Any better idea?

warpech: I think PUT is ok, because this administrative function replaces the current configuraction with the default configuraction. This request is idempotent.

kegor:
@warpech I checked locally, and get same failed tests, I think root cause might be here: https://github.com/Starcounter/Blending/blob/master/test/BlendingEditor.Tests/Test/BaseTest.cs#L77

docs/administration.md
kegor: Ok, if the request is idempotent than it is `almost` PUT :) I agree that this question is not so critical here.

warpech:
I pushed my changes that I thought will fix the problem with the tests. 

There is still some problem, but I cannot debug myself, because when `BlendingEditor_TestHelper` is started, a cURL request `DELETE /blendingeditor/cleardata` causes all apps to crash.

Please check if you can help in any way to make the tests pass.

warpech:
I merged @kegor's fix for the tests that he made on another branch. I think this is good to go. I will merge if @kegor agrees.

PR #271, opened 2018-08-01
fix drag and drop ordering of layout conditions

warpech:
the previous solution only worked for the first drag. Any consecutive drag did not work or was not properly rendered in the client

The changes that I made on the server:
- after reordering reuse the previous values of sort number (just assign them to the desired layout conditions)
- call `GetLayoutConditions` to refresh the list of layout conditions that is sent to the client (this forces the table on the client to re-render the data)
- cancel applying the value to `ReplaceConditions$`, so it remains an empty string that can accept new patches

The changes that I made on the server:
- use a function `resetSorter` to apply table-dragger only once on view load (instead of using Polymers `dom-change` event listeners)
- disable the ability of table-dragger to actually swap `<tr>` elements in DOM using `insertBefore`, because this was causing Polymer `dom-repeat` to go crazy


src/BlendingEditor/Website/ViewModels/AllLayoutConditionsPage.json.cs
```diff
@@ -42,30 +43,22 @@ void Handle(Input.CreateModalTrigger action)
...
+            this.LayoutConditions.Insert(toIndex, condition);
 
-                movedLayoutCondition.SortNumber = toSortNumber;
+            for (var i = 0; i < this.LayoutConditions.Count; i++)
+            {
+                this.LayoutConditions[i].Data.SortNumber = (int)previousSortNumbers[i];
             }
+
+            GetLayoutConditions();
+            action.Cancel();
```
PatrykSzwer: This is not needed. Even if its not empty it will still receive new patch after every change. Please remove this line.

Issue #278, opened 2018-08-14
I cannot change the relevant layout condition of an area item


warpech:
### Steps to reproduce

1. Load the configuration of Omni project in BlendingEditor (https://github.com/Starcounter/HeadsOmni/tree/63efecd0db0b442fb0dea0fb25decd32668c80c8/config)
2. Click on KioskCheckoutLayout
3. Click on Area Items
4. Expand the "Items" button near the "Top Left" area item
5. In the layout condition select `/(shoppingcart|orders).*`
6. Press save

### Expected result

The new layout condition should be saved as the selection for this area item

### Actual result

The "Save" button does not become active after the layout condition is selected

I can cheat that by clicking on the `Disabled` checkbox twice. Then I can save, but still, it does not have any effect to change the saved layout condition.

##### Starcounter version 2.4.990.122

Screen recording: https://drive.google.com/open?id=1C11TJZQ1Z5UBVDqZ2E8Kynd-muye5jP0

kegor:
Closing because related PR https://github.com/Starcounter/Blending/pull/279 was merged.

PR #279, opened 2018-08-14
Problem 278: fix for the Layout Conditions select box on Areas page.

kegor:
**Problem #278:** 
We cannot change the relevant layout condition of an area item.

**Solution:** 
Key field was changed on the Guid in the Layout Conditions select-box and in the related view-model.

**Screenshots/Tests:**
layout_conditions_select_box


src/BlendingEditor/Website/ViewModels/AreasPage.json.cs
```diff
@@ -127,7 +127,7 @@ void Handle(Input.DeleteTrigger action)
 
             void Handle(Input.UrlKey action)
             {
-                this.Data.Url = string.IsNullOrEmpty(action.Value) ? null : Db.SQL<WebUrl>($"select wu from {typeof(WebUrl)} wu where wtg.Guid = ?", action.Value).First();
+                this.Data.Url = string.IsNullOrEmpty(action.Value) ? null : Db.SQL<WebUrl>($"select wu from {typeof(WebUrl)} wu where wu.Guid = ?", action.Value).First();
```
kegor: What do you think about adding Linq compiled query here?

warpech: This is not a performance intensive place, so I don't see a big difference.

src/BlendingEditor/Website/ViewModels/AreasPage.json
```diff
@@ -29,7 +29,7 @@
   ],
   "LayoutConditions": [
     {
-      "Key": "",
+      "Guid": "",
```
warpech: I think we need to reduce our usage of `Guid` to minimum. That means it should be only used when it is needed as an external reference.

In case of communicating between a view-model and a view of the app, the ObjectID/ObjectNo/Key is the correct way to communicate.

Can you rather change the query?

kegor:
Locally unit-tests were ok.

src/BlendingEditor/Website/ViewModels/AreasPage.json
kegor: @warpech I sent iteration with the change in the query and in all related places. The same manual test was successful.

Issue #276, opened 2018-08-08
Export settings process should not be two-step


warpech:
Export settings process right now is two step:

1. You need to click a button to generate a link
2. You need to click the link to download

I find this very annoying, because half of the time I think that I have exported the settings, when in fact I forgot that I need to click on that link.

@PatrykSzwer could you change that so that clicking on the "Export" button directly starts to download?

PatrykSzwer:
Closing as done.

PR #283, opened 2018-08-23
Improve export settings experience

PatrykSzwer:
For https://github.com/Starcounter/Blending/issues/276


src/BlendingEditor/Website/Helpers/MigrationHelper.cs
```diff
@@ -1,17 +1,17 @@
...
+using System.Linq;
 
 namespace BlendingEditor.Website.Helpers
 {
     public class MigrationHelper
     {
-        public static string ExportDataToFile(out string message)
+        public static string ExportDataToFile()
         {
             var filePath = Path.GetTempFileName().Replace(".tmp", ".json");
```
alshakero: [`GetTempFileName`](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?redirectedfrom=MSDN&view=netframework-4.7.2#System_IO_Path_GetTempFileName) creates an empty file in tmp directory. Which means this line will accumulate files with time. Maybe actually rename the file instead of just simply replacing its extension.

src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -14,20 +14,12 @@
...
         <h2 slot="blendingeditor/settings-header">Settings</h2>
         <h3 slot="blendingeditor/settings-database-header">Database</h3>
         <label slot="blendingeditor/settings-export-note">
             Export all configuration data from the current database to a database dump file.<br> The following data will be exported: Layout groups, layouts, areas, area items, layout conditions and compositions.
         </label>
-        <button slot="blendingeditor/settings-export-button" disabled="{{model.ExportInProgress}}" onmousedown="++this.value;" value="{{model.ExportDataTrigger$::click}}">{{model.ExportButtonText}}</button>
-        <template is="dom-if" if="{{model.ExportInProgress}}">
-            <div slot="blendingeditor/settings-export-loader" class="blendingeditor-loader"></div>
-        </template>
+        <a slot="blendingeditor/settings-export-button" href="/blendingeditor/download" download>Export data</a>
```
alshakero: `download` is not needed here. Use `target="_blank"` instead. It is only needed when you can't have attachment header in cases where the server is not yours.

src/BlendingEditor/Website/Api/MainHandlers.cs
```diff
@@ -60,19 +59,16 @@ public void Register()
...
                 {
-                    StreamedBody = stream,
+                    Body = jsonString,
                     ContentType = "application/octet-stream",
                     Headers =
                     {
                         ["content-disposition"] =
-                        $"attachment; filename=Blending_dump_{DateTime.Now:dd/MM/yy-HH:mm:ss}.json",
-                        ["content-length"] = new FileInfo(file).Length.ToString()
+                        $"attachment; filename=Blending_dump_{DateTime.Now:dd/MM/yy-HH:mm:ss}.json"
```
alshakero: I think `/` is illegal to be in file names, can you change it to `-`?

PatrykSzwer: It is saved with dots, like:
`Blending_dump_24.08.18-13_49_43`

PatrykSzwer: But I'll change that to make it clearer

PR #290, opened 2018-09-25
Add formatting for exported json string

PatrykSzwer:
For https://github.com/Starcounter/HeadsOmni/issues/348


warpech:
Tested, works great. Now I get decent diffs when comparing two export files. Still it is not very easy to see what happened inside the HTML string, but that's not the topic of this PR.

diff

PatrykSzwer:
I'm happy to hear that. Merging.

PR #311, opened 2018-10-22
Duplicate layout button

PatrykSzwer:
For https://github.com/Starcounter/Blending/issues/300

Copied layout will have ` - Copy` in its name. Duplicate button will be disabled after first click to prevent multiple duplications by mistake, but after refreshing/changing the page button will still be clickable.


src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -638,6 +638,80 @@ public static WebSettings EnsureSettings()
...
+            var templateCopy = new WebTemplate
+            {
+                Disabled = true,
+                Guid = Guid.NewGuid().ToString(),
+                Html = layout.Html,
+                Name = layout.Name + " - Copy",
+                WebTemplateGroup = layout.WebTemplateGroup
+            };
+
+            foreach (var page in layout.Pages)
```
miyconst: Please no `var` usage when the variable type is not clear.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -638,6 +638,80 @@ public static WebSettings EnsureSettings()
...
+                    Disabled = page.Disabled,
+                    Hidden = page.Hidden,
+                    Url = page.Url,
+                    IsRegex = page.IsRegex,
+                    SortNumber = GetNewWebUrlSortNumber()
+                };
+
+                layoutConditionCopies.Add(page.Guid, layoutConditionCopy);
+
+                foreach (var webHttpHeader in page.Headers)
```
miyconst: Please no `var` usage when the variable type is not clear.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -638,6 +638,80 @@ public static WebSettings EnsureSettings()
...
+                        Guid = Guid.NewGuid().ToString(),
+                        Name = webHttpHeader.Name,
+                        PartialMatching = webHttpHeader.PartialMatching,
+                        Url = layoutConditionCopy,
+                        Value = webHttpHeader.Value
+                    };
+                }
+            }
+
+            foreach (var area in layout.Sections)
```
miyconst: Please no `var` usage when the variable type is not clear.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -638,6 +638,80 @@ public static WebSettings EnsureSettings()
...
+                var areaCopy = new WebSection
+                {
+                    Guid = Guid.NewGuid().ToString(),
+                    Name = area.Name,
+                    Disabled = area.Disabled,
+                    Template = templateCopy,
+                    Default = area.Default
+                };
+
+                foreach (var areaMap in area.Maps)
```
miyconst: Please no `var` usage when the variable type is not clear.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -131,41 +140,41 @@ void Handle(Input.CloseModalTrigger action)
 
         void Handle(Input.SaveModalTrigger action)
         {
-                var notIncludedConfigurations = this.AllThemeConfigurations.Where(x => !x.Include);
-                foreach (var configurationItem in notIncludedConfigurations)
+            var notIncludedConfigurations = this.AllThemeConfigurations.Where(x => !x.Include);
+            foreach (var configurationItem in notIncludedConfigurations)
```
miyconst: Please no `var` usage when the variable type is not clear.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -131,41 +140,41 @@ void Handle(Input.CloseModalTrigger action)
 
         void Handle(Input.SaveModalTrigger action)
         {
-                var notIncludedConfigurations = this.AllThemeConfigurations.Where(x => !x.Include);
-                foreach (var configurationItem in notIncludedConfigurations)
+            var notIncludedConfigurations = this.AllThemeConfigurations.Where(x => !x.Include);
```
miyconst: Please no `var` usage when the variable type is not clear.

src/BlendingEditor/wwwroot/BlendingEditor/views/GeneralPage.html
```diff
@@ -49,6 +52,7 @@ <h2 slot="blendingeditor/general-layout-header">General layout configuration - {
 
     <button slot="blendingeditor/general-save-button" value="{{model.SaveTrigger$::click}}" onmousedown="++this.value;" disabled="{{!model.Trn.IsDirty}}">Save</button>
```
miyconst: Please make sure that all buttons are disabled when clicked to avoid double clicking. The buttons should go back to enabled state when server finishes processing the corresponding trigger handler.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -131,41 +140,41 @@ void Handle(Input.CloseModalTrigger action)
...
+                    relation.Delete();
                 }
+            }
 
-                var configurationsToInclude = this.AllThemeConfigurations.Where(x => x.Include);
-                foreach (var configurationItem in configurationsToInclude)
+            var configurationsToInclude = this.AllThemeConfigurations.Where(x => x.Include);
+            foreach (var configurationItem in configurationsToInclude)
+            {
+                var relation = Db.SQL<CssCollectionRelation>($"SELECT ccr FROM {typeof(CssCollectionRelation)} ccr " +
```
miyconst: Please use `Starcounter.Linq` instead of string concatenation.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -131,41 +140,41 @@ void Handle(Input.CloseModalTrigger action)
...
-                    {
-                        var relation =
-                            Db.SQL<CssCollectionRelation>(
-                                $"SELECT ccr FROM {typeof(CssCollectionRelation)} ccr " +
-                                $"WHERE ccr.{nameof(CssCollectionRelation.Template)} = ? " +
-                                $"AND ccr.{nameof(CssCollectionRelation.Collection)} = ?", this.Data, configurationItem.Data).FirstOrDefault();
-
-                        relation.Delete();
-                    }
+                    var relation =
```
miyconst: Please use `Starcounter.Linq` instead of string concatenation.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
```diff
@@ -131,41 +140,41 @@ void Handle(Input.CloseModalTrigger action)
...
+                        Template = this.Data
+                    };
                 }
+            }
 
-                // Refresh layout theme configurations list
-                this.LayoutThemeConfigurations.Data = Db.SQL<CssCollection>($"SELECT ccr.{nameof(CssCollectionRelation.Collection)} FROM {typeof(CssCollectionRelation)} ccr " +
-                                                                    $"WHERE ccr.{nameof(CssCollectionRelation.Template)} = ?", this.Data);
+            // Refresh layout theme configurations list
+            this.LayoutThemeConfigurations.Data = Db.SQL<CssCollection>($"SELECT ccr.{nameof(CssCollectionRelation.Collection)} FROM {typeof(CssCollectionRelation)} ccr " +
```
miyconst: Please use `Starcounter.Linq` instead of string concatenation.

warpech:
I just discussed with @PatrykSzwer that the custom compositions related to the layout must be duplicated as well.

This is because the main function of a layout is to apply a visual container to a page. Without keepig the same visual container, the duplication is incomplete.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
PatrykSzwer: How in this case It's not clear since the name of the variable is literally the name of the type? But Ok, I'll add it anyway.

src/BlendingEditor/Website/ViewModels/GeneralPage.json.cs
PatrykSzwer: I had to use DbLinq with `.AsEnumerable()` inside here, because without it it was throwing errors regarding the wrong type. Seems like `.Select` is still not working correctly in DbLinq.

PatrykSzwer:
@warpech I added mechanism for copying compositions along with layout. Please check that part.

src/BlendingEditor/Migration/JsonMigrationContainer.cs
```diff
@@ -111,7 +111,7 @@ public void PopulateForExport()
                 .Where(composition => composition.NextVersion == null)
                 .AsEnumerable()
                 .Where(x => IsHtmlViewValid(x.MainView.Uri))
-                .OrderBy(x => HtmlViewCompositionHelper.GetHtmlFileName(x));
+                .OrderBy(HtmlViewCompositionHelper.GetHtmlFileName);
```
warpech: 😍

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -639,6 +639,116 @@ public static WebSettings EnsureSettings()
...
+                };
+
+                var htmlViewCompositions = DbLinq.Objects<HtmlViewComposition>()
+                    .Where(composition => composition.NextVersion == null)
+                    .AsEnumerable()
+                    .Where(composition => composition.MainView?.Equals(htmlView) ?? false);
+
+                foreach (HtmlViewComposition htmlViewComposition in htmlViewCompositions)
+                {
+                    var htmlCompositionCopy = new HtmlViewComposition
```
warpech: Please note that everywhere in this app, new compositions are created using a helper method `CreateFirstOrNextVersion`.

This case should be no exception.

src/BlendingEditor/Website/Helpers/DataHelper.cs
```diff
@@ -639,6 +639,116 @@ public static WebSettings EnsureSettings()
...
+                    };
+
+                    if (areaMap.Url != null)
+                    {
+                        mapCopy.Url = layoutConditionCopies[areaMap.Url.Guid];
+                    }
+                }
+            }
+
+            var htmlView = DbLinq.Objects<HtmlView>().FirstOrDefault(x => x.Uri.Contains(layout.Guid));
```
warpech: This query incorrectly assumes that there is only one `HtmlView` that contains the layout GUID. In fact, there can be more than one. See "KioskLayout" in Omni for example:

```
SELECT * FROM Starcounter.HtmlViewComposition h JOIN Starcounter.HtmlView v ON h.MainView = v WHERE v.Uri LIKE '%4cf8f70f-6a61-495f-9ee9-93c609e2c0a1%'
```

image

PatrykSzwer: Then how exactly should I choose the correct one? How determine the one that should be used for specific composition?

src/BlendingEditor/Website/Helpers/DataHelper.cs
PatrykSzwer: Done.

src/BlendingEditor/Website/Helpers/DataHelper.cs
warpech: Why should you choose to correct one? Since you are copying all areas and area conditions, you should also copy all HTML compositions related to them.

Simple test: duplicate `KioskLayout` in Omni and see if `KioskLayout - copy` renders layout the same on all pages.

PatrykSzwer: I've added mechanism to copy all htmlViews and related to them compositions. Please test again.

warpech:
The code looks good, but I cannot succesfully test it. 

Steps to reproduce:

1. Load Omni config
2. Go to `http://omni.local:8080/showroom/products` to see how it looks
3. Duplicate `KioskLayout`
4. Delete `KioskLayout`
5. Go to `http://omni.local:8080/showroom/products` to see that it looks the same as in point 2

Expected:

The page in step 5 and 2 should look the same

Actual:

In the step 5, I can see an empty page rendered. Administrator shows the following error:

```
System.ObjectDisposedException: The session have been disposed and cannot be used.
```

PatrykSzwer:
@warpech I was trying to test it with Omni but I was receiving empty view + exception on the `/showroom` page, but regardless if it was with or without my copied layout applied. I though that there is something wrong in my configuration of Omni itself (since I wasn't running it for a while) and I tested it with default layouts and custom created composition manually. (also remember that after using duplicate to fully recreate the experience you have to recheck layout conditions order, because newly copied conditions will be at the bottom with highest order numbers).

warpech:
I just tested and it still does not work as expected.

Steps to reproduce:

1. Start Blending app suite on an empty db
2. Log in to BlendingEditor and make sure it is presented in "TopbarLayout"
3. Duplicate the default layout "TopbarLayout" (creates "TopbarLayout - Copy")
4. Delete "TopbarLayout"
5. Reload the page

Expected:

BlendingEditor should be presented in "TopbarLayout - Copy"

Actual:

BlendingEditor is presented without any layout

PatrykSzwer:
Ok, so I was investigating this and seems like something is wrong with deleting layout when there is a copy of it. When you use disable option, it will work correctly, but when you delete copied layout, for me I'm getting no content in the page. I have to investigate how deleting of layout is currently working.

Issue #301, opened 2018-10-02
Save and serve compositions from file system not from database


miyconst:
Currently the blending compositions are saved and distributed through database. It makes it very inconvenient due to:

- Impossible to use any kind of IDE, such as Visual Studio, Visual Studio Code, Notepad++ for editing the compositions.
- Re-creating database resets the compositions.
- Impossible to track versioning of the compositions. It's possible to export compositions into a JSON file, and commit it, but it produces one huge un-formatted file with all blending information in it.
- No easy way to share specified compositions between developers/databases/hosts.
- No easy way to remove/reset certain compositions.
- No easy way to implement and distribute default compositions per app, but outside of the app's HTML views.

That is why I propose to keep and serve the HTML compositions from file system, rather than from the database. Every composition should be presented as a separate file, and the files should be structured into folders for easier navigation and search.

cc @warpech, @tomalec

tomalec:
I totally support above. I also believe the place of HTML is in the HTML file and HTML-in-JSON or HTML-in-C#.

We already discussed it a bit, having the ideas to serve custom compositions from `htmlmerger` so it would be even cleaner for our client-side platform to use it.

Adding HTML compositions to the merged partials would also give use de-duping/performance improvement for free (at least as long as HTML Imports or equivalent will be there)

AFAIR we started to keep **custom** compositions in the DB, as we consider them changeable but our customers, therefore part of customization. Needless to mention legacy from keeping JSON configuration for `juicy-tile-list` or `flexbox-layouts` ;) 👴 .We also planed to have more advanced composition management, versioning, rolling back, iterations between handcrafted ones, and build with machine support (AI blending).

Also as a food for thought before the meeting, I'd like to bring the idea that was also emerging quite often: treating views as data.

I believe none of the above ideas and goals collide. 
We need to find a way to just join them without overcomplicating.

I think we could just introduce convention to keep custom compositions in FS, which could then be fed to DB if needed.

warpech:
Related: https://github.com/Starcounter/Blending/issues/302#issuecomment-426809057

warpech:
We agreed that we will serve compositions from the file system. This allows us the solution owner to install composition dependencies using a package manager.

The compositions will be served from the BlendingProvider's runtime directory.

mmnosek:
We learned that it's hard to work with compositions in Omni project. Please keep in mind that we should make it possible to work on compositions simultaneously. Right now it's a nightmare due to uncomparable `.json` files.

[...]

Issue #234, opened 2018-06-05
Rename LayoutInfo to CompositionInfo


warpech:
Rename the `LayoutInfo` view-model to `CompositionInfo`, because the current name incorrectly describes the purpose of that class.

warpech:
Implemented in #320

PR #320, opened 2018-11-01
Edit compositions in files

warpech:
Implements #301, #234 

I tried to make it not complex but it is a significant change anyway. The code changes are explained in the commit messages. 

The feature from the solution owner's perspective is explained in the UI of BlendingEditor. I will not give any clues on how to use it, in order to get more genuine feedback from people who try using it :)

The proposed configuration for the Omni project is in https://github.com/Starcounter/HeadsOmni/tree/use-composition-folder

The architecture of this solution is presented on the diagram:

image


miyconst:
Making a thoughtful review of the code is going to take some time, but here are a few high level items:

- Currently the composition file name contains information about all attached views, which is limited [to 255 chars](https://stackoverflow.com/questions/265769/maximum-filename-length-in-ntfs-windows-xp-and-windows-vista). In my case, running just a few app resulted in a file name of ~100 chars long. 
- The solution does not watch for the file changes during runtime, and does not update the database, it's a must have feature.
- In my case, my files were overwritten by the database composition values on application startup, instead of picking up the file changes into the database. I can help debugging if needed.

I do not have an ideal solution for the file names with attached views, but here are a few things:

- It would be nice to have the folder structure the same as the main view. So `/_People_viewmodels_PhoneNumberPage.html` would be `/People/viewmodels/PhoneNumberPage.html`.
- In my solution the information about attached views is saved inside the file itself. `<!-- [ "/bim-projects/project/project-page.html", "/Buildcraft/ViewModels/ProjectTabsPage.html" ] -->`. It limits the number of compositions per main view to one, but serves the purpose well, since there were no need (for now) to run different set of app.

warpech:
> Currently the composition file name contains information about all attached views, which is limited to 255 chars. In my case, running just a few app resulted in a file name of ~100 chars long.

I am aware. I tried to follow the [new .Net way of removing the path length limit](https://blogs.msdn.microsoft.com/jeremykuhne/2016/07/30/net-4-6-2-and-long-paths-on-windows-10/), but I failed at that.

For now, I have implemented a fancy file naming function that truncates the attached view names after 20 characters. All custom compositions used in Omni fit within the limit now.

> The solution does not watch for the file changes during runtime, and does not update the database, it's a must have feature.

It does not watch but it picks the new content when you load a page that uses the composition. It was a 5 minute solution. I have no experience in implementing file observers in .Net and would be happy not to do that in this PR.

> In my case, my files were overwritten by the database composition values on application startup, instead of picking up the file changes into the database. I can help debugging if needed.

Can you please give steps to reproduce?

> It would be nice to have the folder structure the same as the main view. So `/_People_viewmodels_PhoneNumberPage.html` would be `/People/viewmodels/PhoneNumberPage.html`.

Interesting idea, but I am not sure if I like it. I kind of like having a flat structure. We must decide early on that because any changes to the file structure will be a breaking change. @PatrykSzwer, @alshakero do you support flat or deep folder structure?

> In my solution the information about attached views is saved inside the file itself. `<!-- [ "/bim-projects/project/project-page.html", "/Buildcraft/ViewModels/ProjectTabsPage.html" ] -->`. It limits the number of compositions per main view to one, but serves the purpose well, since there were no need (for now) to run different set of app.

I wanted a solution that is 100% compatible with the current data model, which allows multiple combinations of main view and attached views. This though is a subject of a separate issue: https://github.com/Starcounter/HeadsOmni/issues/604

miyconst:
>I am aware. I tried to follow the new .Net way of removing the path length limit, but I failed at that.

I think, the path length limit and the file name length limit are different things, and removing path length limit does not remove the file name length limit. Please correct me if I am wrong.

miyconst:
>Can you please give steps to reproduce?

In my case it was:

- Start apps.
- Open generated file.
- Make a change.
- Save file.
- Reload page - no changes visible.
- Restart the apps - the file is reverted to the database version of the composition.

miyconst:
>Interesting idea, but I am not sure if I like it. I kind of like having a flat structure.

Just to note: large number of files in a folder may cause performance issues when listing files in the folder.

alshakero:
I'm happy about this. I edited the composition and could see my changes on disc and diffable quite nicely. And I wouldn't mind it merged as in. Tested a few views and it works. 

One more wish I have: Is it feasible to remove the whole DB part? I feel that it's over complicated right now. When I run the app it's writing files on disk without me knowing, then when I edit the files directly I see no effect (as @miyconst mentioned). And if I edit/delete those files, then edit in composition editor, my changes to the files are discarded. I'm kinda wary of the double source of truth method. Keeping it all in files would save us from a lot of bugs.

warpech:
I pushed some changes. Now it observes the files correctly and informs explicitly that the files in the folder might be overwritten.

PatrykSzwer:
There are multiple conflicts in this PR, please fix.

src/BlendingCommon/HtmlViewCompositionHelper.cs
```diff
@@ -91,5 +92,81 @@ public static HtmlView GetView(string viewUri)
         {
             return Db.SQL<HtmlView>("SELECT v FROM Starcounter.HtmlView v WHERE v.Uri = ?", viewUri).FirstOrDefault();
```
PatrykSzwer: Please use starcounter.Linq instead of SQL query.

src/BlendingCommon/HtmlViewCompositionHelper.cs
```diff
@@ -91,5 +92,81 @@ public static HtmlView GetView(string viewUri)
...
+            }
+
+            fileName += ".html";
+
+            return GetSafeFilename(fileName);
+        }
+
+        public static string[] GetViewUris(HtmlViewComposition composition)
+        {
+            return composition.Attachments.Select(x => x.AttachedView.Uri).Prepend(composition.MainView.Uri).ToArray();
```
PatrykSzwer: I'm getting here and under that multiple VS errors about referencing `BindingInterfaces`:
image

I saw that in Blending app before. Referencing it solves the problem but even without that It builds without errors. @miyconst maybe you know the reason?

src/BlendingEditor/Composition/Api/MappingHandlers.cs
```diff
@@ -7,6 +8,21 @@ internal class MappingHandlers
         public void Register()
         {
             Handle.GET("/blendingeditor/compositioninserter", () => new CompositionInserter());
+
+            Hook<HtmlViewComposition>.CommitInsert += (s, a) =>
+            {
+                this.SaveToFile(a);
```
PatrykSzwer: "a" says nothing, please rename to `composition` or something like that.

src/BlendingEditor/Composition/Api/MappingHandlers.cs
```diff
@@ -7,6 +8,21 @@ internal class MappingHandlers
...
+            };
+
+            Hook<HtmlViewComposition>.BeforeDelete += (s, a) =>
+            {
+                Helpers.FileSystemHelper.DeleteFile(a);
```
PatrykSzwer: Same as above.

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+using BlendingCommon;
+using BlendingEditor.Website.Helpers;
+
+namespace BlendingEditor.Composition.Helpers
+{
+    public class FileSystemHelper
+    {
+        public static string RegisteredCompositionsFolder;
+
+        private static FileSystemWatcher Watcher;
```
PatrykSzwer: Private static field like that should be name `_watcher`.

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+
+            return null;
+        }
+
+        public static string GetCompositionsFolder()
+        {
+            var settings = DataHelper.GetSettings();
+            if (settings?.Description != null && settings?.Description != String.Empty)  //TODO rename Description to CompositionsFolder
```
PatrykSzwer: Can be simplified and replaced with:
`            return !string.IsNullOrEmpty(settings?.Description) ? settings.Description : "";`

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+            }
+
+            return "";
+        }
+
+        public static string GetAbsoluteFolderPath(string relativePath)
+        {
+            var rootPath = Path.GetDirectoryName(Application.Current.FilePath);
+
+            if (relativePath != string.Empty) return Path.GetFullPath(Path.Combine(rootPath, relativePath));
```
PatrykSzwer: Can be simplified to:
`            return relativePath != string.Empty ? Path.GetFullPath(Path.Combine(rootPath, relativePath)) : null;`

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -0,0 +1,209 @@
...
+
+            return null;
+        }
+
+        public static void RegisterSharedFolder()
+        {
+            if (Watcher != null) Watcher.Dispose();
```
PatrykSzwer: Use null propagation (` Watcher?.Dispose();`)

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
```diff
@@ -1,9 +1,42 @@
+using System.Linq;
 using Starcounter;
+using BlendingCommon;
+using BlendingEditor.Composition.Helpers;
 
 namespace BlendingEditor.Website.ViewModels
 {
     partial class CompositionPage : Json, IExplicitBound<HtmlViewComposition>
     {
+        private string OriginalHtml;
```
PatrykSzwer: Should be lowercase - `originalHtml`

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -1,13 +1,49 @@
...
 
 namespace BlendingEditor.Website.ViewModels
 {
     partial class SettingsPage : Json
     {
         public string SessionId => Session.Current.SessionId;
+
+        public string CompositionsFolder => DataHelper.GetSettings()?.Description; //TODO rename Description to CompositionsFolder
+
+        public string RegisteredCompositionsFolderFullPath => FileSystemHelper.GetAbsoluteFolderPath(Composition.Helpers.FileSystemHelper.RegisteredCompositionsFolder);
```
PatrykSzwer: Full namespace is redundant here, `FileSystemHelper.RegisteredCompositionsFolder` should be enough

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -1,13 +1,49 @@
...
+
+        public string CompositionsFolderFullPath => FileSystemHelper.GetAbsoluteFolderPath(Composition.Helpers.FileSystemHelper.GetCompositionsFolder());
```
PatrykSzwer: Same as above.

src/BlendingProvider/Composition/ViewModels/CompositionInfo.json.cs
```diff
@@ -0,0 +1,43 @@
...
+        private void Handle(Input.Composition input)
+        {
+            var newComposition = Save(input.Value);
+            FillInCompositionInfo(newComposition);
+        }
+
+        public HtmlViewComposition Save(string html)
+        {
+            HtmlViewComposition newComposition = null;
+            Starcounter.Internal.StarcounterEnvironment.RunWithinApplication(Program.AppNameForDataAccess, () => {
```
PatrykSzwer: You're using obsolete methods here. I understand it can be necessary but it would be good if you would provide some short comment about why It is needed here.

src/BlendingProvider/Composition/ViewModels/CompositionInfo.json.cs
```diff
@@ -0,0 +1,43 @@
...
+                    var viewUris = PartialId.Split('|');
+                    newComposition = HtmlViewCompositionHelper.CreateFirstOrNextVersion(viewUris, html);
+                });
+            });
+            return newComposition;
+        }
+
+        public void FillInCompositionInfo(HtmlViewComposition composition)
+        {
+            Starcounter.Internal.StarcounterEnvironment.RunWithinApplication(Program.AppNameForDataAccess, () => {
```
PatrykSzwer: Same as above.

src/BlendingCommon/HtmlViewCompositionHelper.cs
warpech: This is not my code, it is just refactored by moving from one class to another. Frankly speaking, it is high time to remove that legacy migration method.

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
warpech: Not according to the style guide: https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/coding_style/csharp_code_formatting.md

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
warpech: Not according to the style guide: https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/coding_style/csharp_code_formatting.md

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
miyconst: This line fails with the following exception when a wrong path entered:

```
HResult=-2146233088
HelpLink=https://docs.starcounter.io/?q=SCERR14012
---> System.NotSupportedException: The given path's format is not supported.
   at System.Security.Permissions.FileIOPermission.EmulateFileIOPermissionChecks(String fullPath)
   at System.Security.Permissions.FileIOPermission.QuickDemand(FileIOPermissionAccess access, String fullPath, Boolean checkForDuplicates, Boolean needFullPath)
   at System.IO.Path.GetFullPath(String path)
   at BlendingEditor.Composition.Helpers.FileSystemHelper.GetAbsoluteFolderPath(String relativePath) in C:\Users\Miyconst\Documents\GitHub\Buildcraft\src\Blending.Mapper\Blending\src\BlendingEditor\Composition\Helpers\FileSystemHelper.cs:line 49
   at BlendingEditor.Website.ViewModels.SettingsPage.get_CompositionsFolderFullPath() in C:\Users\Miyconst\Documents\GitHub\Buildcraft\src\Blending.Mapper\Blending\src\BlendingEditor\Website\ViewModels\SettingsPage.json.cs:line 20
   at BlendingEditor.Website.ViewModels.SettingsPage.get_CompositionsFolderStatus() in C:\Users\Miyconst\Documents\GitHub\Buildcraft\src\Blending.Mapper\Blending\src\BlendingEditor\Website\ViewModels\SettingsPage.json.cs:line 32
   at lambda_method(Closure , Json )
HResult=-2146233067
```

src/BlendingCommon/HtmlViewCompositionHelper.cs
```diff
@@ -0,0 +1,104 @@
...
+
+            if (fileName == null)
+            {
+                return null;
+            }
+
+            fileName = GetSafeFilename(fileName);
+            fileName = Path.GetFileNameWithoutExtension(fileName);
+
+            foreach (var attachment in composition.Attachments)
```
miyconst: The SQL `SELECT` result is never guaranteed to be in any specific order without `ORDER BY`, thus the `GetHtmlFileName` method is producing different results for the same composition.

```suggestion
            foreach (var attachment in composition.Attachments.OrderBy(x => x.AttachedView.Uri))
```

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
miyconst: The guide is outdated, and the apps team in form of @LarsTjernstrom decided to use `_` prefix for the private members.

I do not support this, for me `this.foo` is much better than `_foo`, but majority seems to like `_foo` instead, so I am ready to adapt.

In this particular case I would rather propose `public string OriginalHtml { get; protected set; }`, which would satisfy both, @warpech and @PatrykSzwer 😄.

Anyway, it's probably time to revisit our coding guides.

src/BlendingCommon/HtmlViewCompositionHelper.cs
warpech: Wow, I didn't know that!

In your suggestion, you are letting Linq `OrderedEnumerable` sort the result set instead of Starcounter database engine. Isn't it preferable to add the `ORDER` clause within the query?

miyconst: That's right, it's better to sort it on the database level. The proper way would be to change `HtmlViewComposition.Attachments` property:

```diff
-public IEnumerable<HtmlViewAttachment> Attachments => Db.SQL<HtmlViewAttachment>("SELECT a FROM Starcounter.HtmlViewAttachment a WHERE a.Composition = ?", this);
+public IQueryable<HtmlViewAttachment> Attachments => DbLinq.Objects<HtmlViewAttachment>().Where(x => x.Composition == this);
```

But since we don't want to depend on [`Starcounter.Linq`](https://github.com/Starcounter/Starcounter.Linq) in Level1, the easy solution is to use full query instead of the navigation property:

```cs
DbLinq.Objects<HtmlViewAttachment>().Where(x => x.Composition == Composition).OrderBy(x => x.AttachedView.Uri);
```

miyconst: P.S. In this particular case, the performance of the database and the .NET CLR sorting will be the same.

LarsTjernstrom:
@miyconst , @warpech 

Since constructur injection was not used by Starcounter developers, there was no need to name a private property with underscore. 
Now that we use constructur injection this start being a "problem". 
When googling what is reccomended...it turns out that even Microsoft themselves in new libraries use undercore for private properties! So even the "Gods" don't always 
```
this.Logger = logger

but the write

_logger = logger
```




Note, that codig guildelines are far more less important than design guidelines, since it's in the big picture you make the big misstakes.

miyconst:
@LarsTjernstrom I see your point, and as I already said, I am willing to adjust 👍.

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
warpech: Sorry, that was clearly a typo in the guideline because that point is about underscore. I have updated the guideline.

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
warpech: I have nothing agaist new guidelines, but please advertise the correct ones throughout the team. I will not apply them here because of anectodal referrence.

src/BlendingCommon/HtmlViewCompositionHelper.cs
warpech: The method was removed.

src/BlendingProvider/Composition/ViewModels/CompositionInfo.json.cs
warpech: Comment added, please check.

src/BlendingCommon/HtmlViewCompositionHelper.cs
warpech: Thanks! I have appllied your advice in the commit https://github.com/Starcounter/Blending/pull/320/commits/af320e2214c255fdd489e0de19f164c13e24e154.

To apply it in Starcounter, I would need to change `Level1` and I decided not to.

warpech:
Thank you for your review on code style, but I reject to make any guideline changes in this PR other than these that are according to https://github.com/Starcounter/CompanyTrack/blob/master/AppsTeam/Guidelines/coding_style/csharp_code_formatting.md.

I am all for improving the code style guide (or even better switching to Microsoft's one), but not in this PR. This PR is urgent and I don't want to spill th ebaby with the bathwater by discussing coding style in it.

I applied other advice. Please re-review.

miyconst:
@warpech please also fix the [FileSystemHelper.GetAbsoluteFolderPath](https://github.com/Starcounter/Blending/pull/320/files#r230670156) method, and then it's okay to merge.

warpech:
> @warpech please also fix the FileSystemHelper.GetAbsoluteFolderPath method, and then it's okay to merge.

I totally missed that comment, because GitHub does not show it in the "Conversation" tab of this PR for some reason 😮

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
warpech: I totally missed this comment, it is not on the "Conversations" tab 😮

warpech:
After this is merged:

- [ ] update related issues
- [ ] make an issue for the mapping team to rename `Description` to `CompositionsFolder`

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -0,0 +1,215 @@
...
+            var rootPath = Path.GetDirectoryName(Application.Current.FilePath);
+
+            if (relativePath != string.Empty)
+            {
+                var folderPath = Path.Combine(rootPath, relativePath);
+                try
+                {
+                    return Path.GetFullPath(folderPath);
+                }
+                catch (Exception e)
```
miyconst:
```suggestion
                catch (Exception)
```

src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -0,0 +1,215 @@
...
+        {
+            var settings = DataHelper.GetSettings();
+            return !string.IsNullOrEmpty(settings?.Description) ? settings.Description : ""; //TODO rename Description to CompositionsFolder
+        }
+
+        public static string GetAbsoluteFolderPath(string relativePath)
+        {
+            var rootPath = Path.GetDirectoryName(Application.Current.FilePath);
+
+            if (relativePath != string.Empty)
```
miyconst:
```suggestion
            if (!string.IsNullOrWhiteSpace(relativePath))
```

warpech:
@PatrykSzwer that exception is fixed now. Please re-test

warpech:
The automated tests and my manual checks passed, merging

miyconst:
Well done @warpech, as always.

PR #322, opened 2018-11-01
Prototype solution to notifying user when he wants to leave a dirty view-model

warpech:
This is to address: https://github.com/Starcounter/HeadsOmni/issues/250

this code is based on suggestions by @miyconst and @alshakero

I am not very pleased with the solution, because:

1. I would like to know directly in the view-model of TransactionOverviewPage if any view-model in the session is dirty. This would be also a good basis for the Save Bar feature. I could not find a way to do it.
2. Because of (1), I needed to perform duck type search of dirty viewmodels by searching for nodes in the JSON tree with the key "IsDirty" and the value "true". That is very bad but works for the known apps
3. Clicking on a link does no neccessarily mean that a dirty view-model will be detached. It might be a false-positive, when there indeed there is some navigation, but the dirty view-model is in the part of the screen that remains.
4. Clicking on a link does not mean that the dirty view-model changes will actually be not saved. Depending on the app logic, they still may or may not me saved. I had a big problem about that to write in the modal dialog, because saying that changes will be abandoned is a huge speculation actually. Maybe we will make the user think that he indeed abandoned the changes, but still the changes will be saved. That sucks.
5. Clicking on a link is not the only way in which a dirty view-model is abandoned. This can also happen based on any other interaction, which will be undetected by the method used here.

The correct solution will use https://github.com/Starcounter/uniform/issues/203, and this workaround is in my opinion 70% bad, but here it is.

@alshakero, @miyconst perhaps you'd like to review from the technical perspective. Maybe you have better suggestions how to address the problems listed.

@mmnosek, @stefanedqvist perhaps you'd like to review from the UX perspective. Please consider the tradeoffs listed above.

image


warpech:
If, despite being 70% bad, you'd like to see this solution used in Omni, I suggest to not merge this PR but copy the solution to a custom composition of the backoffice layout.

This PR is just to create an easy way to try it out. It works in all apps, even though being provided by just one app (BlendingProvider in this case).

warpech:
Frankly @alshakero's proposed solution is a little bit better (just 51% bad), because it allows pages to actively prevent navigation at the expense of more work for the app authors: https://starcounter.slack.com/archives/GD7DL4JU9/p1540892760007600?thread_ts=1540844898.006000&cid=GD7DL4JU9

If you wish, I can try implementing it.

A 100% good solution is the one that uses server-side logic to determine loosing data in a dirty view-model.

src/BlendingProvider/wwwroot/BlendingProvider/views/TransactionOverviewPage.html
```diff
@@ -0,0 +1,58 @@
...
+                    });
+                }
+                else {
+                    baseFn.call(this, ...arguments);
+                }
+            };
+
+            function recursiveFindDirty(obj) {
+                var dirties = [];
+                if (obj["IsDirty"] != undefined) {
```
alshakero: I would do 

```js
                if (obj.IsDirty) {
                	return true;
                }
```

And save recursing deeper, an array, and another loop to look for a true value in the array.

src/BlendingProvider/wwwroot/BlendingProvider/views/TransactionOverviewPage.html
```diff
@@ -0,0 +1,58 @@
...
+                    dirties.push(obj)
+                }
+                var keys = Object.keys(obj);
+                keys.forEach((key) => {
+                    if (obj[key] === Object(obj[key])) {
+                        dirties = dirties.concat(recursiveFindDirty(obj[key]));
+                    }
+                });
+                return dirties;
```
alshakero: And return `false` here.

src/BlendingProvider/wwwroot/BlendingProvider/views/TransactionOverviewPage.html
```diff
@@ -0,0 +1,58 @@
...
+        (function () {
+            const template = document.currentScript.previousElementSibling;
+            const palindrom = document.querySelector("palindrom-client").palindrom;
+            const baseFn = palindrom.morphUrl;
+
+            template.continueNavigation = null;
+
+            palindrom.morphUrl = function () {
+                var dirties = recursiveFindDirty(this.obj);
+                var isDirty = dirties.find(x => x.IsDirty === true);
```
alshakero: This is not needed if my suggestion below is right.

warpech:
Thank you @alshakero for the review. I will refrain from updates until @mmnosek and @stefanedqvist confirm that it addresses the original problem good enough.

alshakero:
Also, I would use a native `confirm`. It's accessible for free, synchronous, and users are used to it.

Also, one objective benefit is that we'll also need to cover when the user navigates away from our app (a non-morphable navigation). Like typing a URL or refreshing the page. There is `onbeforeunload` event that allows us to show them a question whether they're sure, and this one requires using native `confirm` for sure.

miyconst:
@alshakero has provided good feedback, I have nothing more to add.

mmnosek:
> I would like to know directly in the view-model of TransactionOverviewPage if any view-model in the session is dirty. This would be also a good basis for the Save Bar feature. I could not find a way to do it.

For the Save Bar, we can think about adding Starcounter API for this.

> Because of (1), I needed to perform duck type search of dirty viewmodels by searching for nodes in the JSON tree with the key "IsDirty" and the value "true". That is very bad but works for the known apps

Currently, by default, all view models share the same long-running transaction scope. Can't we just assume, that just if `TransactionOverviewPage` transaction is dirty, we show this dialog? I know it's not ideal, but to make Save/Cancel button work, we rely on this anyway. And Save/Cancel affect only "shared" transaction anyway.

> Clicking on a link does no neccessarily mean that a dirty view-model will be detached. It might be a false-positive, when there indeed there is some navigation, but the dirty view-model is in the part of the screen that remains.

I understand, but in Omni I don't see any place like this (I am not 100% sure).

> Clicking on a link does not mean that the dirty view-model changes will actually be not saved. Depending on the app logic, they still may or may not me saved. I had a big problem about that to write in the modal dialog, because saying that changes will be abandoned is a huge speculation actually. Maybe we will make the user think that he indeed abandoned the changes, but still the changes will be saved. That sucks.

Again. We relay on the fact that all view models share transactions. Not ideal, but enough for Omni. Speaking about UX, I think it's correct. This is how it works on pages like facebook.

> Clicking on a link is not the only way in which a dirty view-model is abandoned. This can also happen based on any other interaction, which will be undetected by the method used here.

We have to live with this.

To sum up, I think we should go for it! I don't think that we should do it on a specific application level. It is not a 100% correct solution anyway.

warpech:
> Currently, by default, all view models share the same long-running transaction scope. Can't we just assume, that just if `TransactionOverviewPage` transaction is dirty, we show this dialog? I know it's not ideal, but to make Save/Cancel button work, we rely on this anyway. And Save/Cancel affect only "shared" transaction anyway.

That would be my thinking. too.

The problem is that in C#, `IsDirty` is always false. I wrote about that in point 1 of the OT. Perhaps it is because C# code executes in the response middleware and has no way to attach to an already existing scope. From @miyconst's response, I understand that he looked at it and could not provide a solution.

----

Regarding the UI, I will switch to the native `confirm` as advised.

warpech:
I am not sure about the native `confirm`. @mmnosek can you make the desicion? 

Despite both solutions block the user interaction, the most significant difference is that `paper-dialog` actually has a modality courtain.

`paper-dialog`:

2018-11-05 16_16_53-window

Native `confirm`:

2018-11-05 16_20_43-window

mmnosek:
I like `paper-dialog` much more 😄

alshakero:
The issue with paper dialog is that it won't work for real navigation. Isn't this a deal breaker?

warpech:
By "real navigation", I guess @alshakero means navigating away from the solution by clicking a link with a `target` attribute or a link to another domain.

This PR does not event attempt observing for real navigation, should it? The issue https://github.com/Starcounter/HeadsOmni/issues/250 does not mention it.

mmnosek:
So the question - is clicking "back" in the browser a real navigation or it's covered by this implementation?

warpech:
> So the question - is clicking "back" in the browser a real navigation or it's covered by this implementation?

If the back button takes you from one page in HeadsOne to another page in Heads One in the same session, it is covered.

mmnosek:
@warpech can we merge this and move this forward?

warpech:
@miyconst I managed to implement server-side check for dirty view-models using JSON merger callback + server-side dirty checker + session store.

The implementation is in https://github.com/Starcounter/Blending/pull/322/commits/3b3744e9ede483be778f166ac1f9020956faf97b

WDYT? I think such solution not only could be merged but also be the basis for the poor man's Save Bar implementation.

mmnosek:
> WDYT? I think such solution not only could be merged but also be the basis for the poor man's Save Bar implementation.

@warpech I suggest we should definitely merge it as it is.

warpech:
I need to discuss the current solution with Konstantin to know that this solution is something that works by design, not by accident. Meanwhile, I think it is ready for testing, so I recomment @mmnosek that you give it a try by:

1. Checkout https://github.com/Starcounter/Blending/tree/confirm-navigation
2. Uncomment the comments in the composition part of https://github.com/Starcounter/Blending/blob/confirm-navigation/src/BlendingProvider/wwwroot/BlendingProvider/views/TransactionOverviewPage.html

warpech:
I just explained the solution over a web meeting to @miyconst. He will write his opinion here.

My planned next steps are:

- merge this PR as is, with enabled Navigation Confirmation Popup and disabled Save Bar by default
- next week:
   - implement Palindrom changes as events instead of monkey patching that is used now
   - add tests in BlendingProvider
   - describe the solution in README.md

@mmnosek can you pls verify Navigation Confirmation Popup with Heads Omni before I merge? It is possible now.

src/BlendingProvider/Composition/ViewModels/CompositionInfo.json.cs
```diff
@@ -2,11 +2,35 @@
...
 namespace BlendingProvider.Composition.ViewModels
 {
     partial class CompositionInfo : Json
     {
+        public bool IsCurrentScopeDirty
+        {
+            get
+            {
+                var currentValue = false;
+                if (this.AttachedScope != null)
```
miyconst: Please remove redundant nesting level.

```suggestion
                if (this.AttachedScope == null)
                {
                    return false;
                }
```

src/BlendingProvider/Website/ViewModels/TransactionOverviewPage.json
```diff
@@ -0,0 +1,6 @@
+{
+  "Html": "/blendingprovider/views/TransactionOverviewPage.html",
+  "IsAnyScopeDirty": false,
+  "SaveAnyScopeTrigger$": 0,
```
miyconst:
```suggestion
  "SaveEveryScopeTrigger$": 0,
```

src/BlendingProvider/Website/ViewModels/TransactionOverviewPage.json
```diff
@@ -0,0 +1,6 @@
...
+  "CancelAnyScopeTrigger$": 0
```
miyconst:
```suggestion
  "CancelEveryScopeTrigger$": 0
```

warpech:
Thanks @miyconst. All done

mmnosek:
> If the back button takes you from one page in HeadsOne to another page in Heads One in the same session, it is covered.

For me, it doesn't work. Other than this works well!

warpech:
>> If the back button takes you from one page in HeadsOne to another page in Heads One in the same session, it is covered.

> For me, it doesn't work. Other than this works well!

Thanks for testing this! This is definitely true. Moreover, it will require a fix in Palindrom that will take few days...

All I can recommend now is to release this PR as is and work on a proper solution next week. Do you agree?

mmnosek:
Agree! Let's merge and then do additional improvements.

warpech:
Great! Made a new issue: https://github.com/Starcounter/HeadsOmni/issues/new

PR #323, opened 2018-11-07
Adding App suffix to application targets

ubbeK:


ubbeK:
These changes have nothing todo with Blending app or blending builds themselves. They are needed for our internal projects and their builds.

@warpech @PatrykSzwer

PR #326, opened 2018-11-08
Natural reference in export

warpech:
Use natural numbers instead of object IDs for export of custom compositions. This implements point 3 of https://github.com/Starcounter/Blending/issues/301#issuecomment-436997947.

The recommended way to test is to:

1. checkout https://github.com/Starcounter/HeadsOmni/blob/natural-reference-in-export/config/blendingeditor.json (the file that I exported from my build of this PR)
2. import that file
3. export that file
4. compare if the exported file is identical than the imported one


warpech:
Thanks for checking! I think I know what the problem was. My last commit should fix that. Merging. Please report or try to investigate any unexpected diffs that you encounter from now on. Comparing two exports should no longer give differences in ids, unless some objects were added or removed.

PR #331, opened 2018-11-14
JSON export/import stabilisation

warpech:
Many improvements for JSON export/import that are needed for stable implementation of https://github.com/Starcounter/Blending/issues/301.

Most importantly:

- In the export file, `Html` value is `null` if the compositions folder is set
- The order of compositions related data in the export file is no longer sorted by ObjectNo. Now it is sorted by something closest to `Guid`: the filename of the composition
- JSON export/import should now be completely stable, meaning that it gives the same result for each developer that uses it


warpech:
For easier verification, this file contains the current Omni configuration imported then exported using this PR: https://github.com/Starcounter/HeadsOmni/blob/aa3a7c4c2ab6d3f33305a0c8f3d399f2518fa51a/config/blendingeditor.json

src/BlendingEditor/Composition/Api/MappingHandlers.cs
```diff
@@ -13,16 +14,14 @@ public void Register()
             {
                 this.SaveToFile(composition);
             };
-
-            Hook<HtmlViewComposition>.BeforeDelete += (s, composition) =>
```
PatrykSzwer: Since you deleted this commit hook I would suggest adding method for deleting composition file under the `Delete` button on compositions page. That way we would avoid piles of files that are not used anymore and are in composition folder for no reason. You can create separate issue for that.

warpech:
Tests passed, merging

Issue #334, opened 2018-11-15
Exception when the observed folder is deleted on disk


warpech:
Copied from https://github.com/Starcounter/Blending/pull/331#pullrequestreview-175305135:

---

Code looks fine except one thing that I mentioned below.

One problem though. I wanted to test this `Recreate files from database` feature and I accidentally found a bug. When you delete the compositions folder and try to use it you will recieve an exception: 
```
starcounter.html:26 PalindromConnectionError: Server error
	System.IO.DirectoryNotFoundException: Could not find a part of the path 'D:\repositories\HeadsOmni\config\compositions\_blendingprovider_blender_layout_4cf8f70f-6a61-495f-9ee9-93c609e2c0a1_area_b0619532-5fec-4845-9019-1e26ca57a052_uri_8e7c14f5-5450-49ac-8cb5-7d28a56cd519+HomeIconPage.html'.
	   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
	   at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
	   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
	   at System.IO.StreamWriter.CreateFile(String path, Boolean append, Boolean checkHost)
	   at System.IO.StreamWriter..ctor(String path, Boolean append, Encoding encoding, Int32 bufferSize, Boolean checkHost)
	   at System.IO.File.InternalWriteAllText(String path, String contents, Encoding encoding, Boolean checkHost)
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.WriteToFileIfDifferent(String path, String content, Boolean force)
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.CopyFromCacheToFile(HtmlViewComposition composition, Boolean force)
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.CopyFromCacheToFiles()
	   at BlendingEditor.Website.ViewModels.SettingsPage.Handle(RecreateFilesFromDb action)
	   at lambda_method(Closure , Json , Input`1 )
[...]

PR #364, opened 2018-11-28
Add check for existance of the compositions folder 

PatrykSzwer:
For #334


warpech:
One test failed in the build https://teamcity.starcounter.org/viewLog.html?buildId=80806&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24. Might it be flaky?

I have started another build.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -31,6 +31,11 @@ public string CompositionsFolderStatus
                 var compositionsFolder = ConvertNullToEmptyString(CompositionsFolder);
                 var registeredCompositionsFolder = ConvertNullToEmptyString(FileSystemHelper.RegisteredCompositionsFolder);
 
+                if (!Directory.Exists(CompositionsFolderFullPath))
+                {
+                    return "COMPOSITION_FOLDER_BAD_VALUE";
+                }
```
warpech: This makes line 45 redundant.

warpech:
Again the build failed with another test that appears flaky: https://teamcity.starcounter.org/viewLog.html?buildId=81021&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24

I started a new build: https://teamcity.starcounter.org/viewLog.html?buildId=81122&tab=buildResultsDiv&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24

And made an issue to investigate flaky tests: https://github.com/Starcounter/Blending/issues/375

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -142,7 +146,14 @@ void Handle(Input.CompositionsFolderApplyNow action)
 
         void Handle(Input.RecreateFilesFromDb action)
         {
-            if(!string.IsNullOrEmpty(FileSystemHelper.RegisteredCompositionsFolder))
+            if (!File.Exists(FileSystemHelper.RegisteredCompositionsFolder))
+            {
+                // Invoke this method to recalculate CompositionsFolderStatus value and send patch to the client
+                Session.Current.CalculatePatchAndPushOnWebSocket();
+                return;
+            }
```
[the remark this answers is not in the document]
warpech: Regarding 2, @PatrykSzwer explained to me on Slack that it would require to watch the *parent* folder, which might be Desktop or C:\. That might be ineffective. I must agree. Alternatively, we could at least check for the folder existance at every write.

PatrykSzwer:
>I can still make an uncaught exception if I:

I pushed fix for that, but its not completed yet due to fact that I want to add some message returned on Composition page when folder is missing. I didn't do that because I also requested adding message mechanism to that page here: https://github.com/Starcounter/Blending/pull/372 and I don't want to do the same thing and create conflicts. When https://github.com/Starcounter/Blending/pull/372 will be merged I'll fix that part.

warpech:
I think that the problem explained above is fixable by checking `IsRegisteredCompositionsFolder` in line https://github.com/Starcounter/Blending/blob/e30fa0c35f67a149e0766d0327f1b40f2170b837/src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs#L39

If you agree, let's do it in this PR.

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
```diff
@@ -67,6 +67,11 @@ protected override void OnData()
 
         void Handle(Input.SaveTrigger action)
         {
+            if (!Directory.Exists(FileSystemHelper.RegisteredCompositionsFolder))
```
warpech: Everywhere else we use `FileSystemHelper.IsRegisteredCompositionsFolder` helper property for that.

Could you please make it DRY, that is either change the implementation of `IsRegisteredCompositionsFolder` or switch to `Directory.Exists(FileSystemHelper.RegisteredCompositionsFolder)` everywhere else?

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
warpech: There is something wrong with the status now. It shows an error message when the field is empty. That should not be an error, because it means you don't want to store compositions in files.

2018-12-04 14_42_27-blendingeditor

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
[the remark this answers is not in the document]
warpech: @PatrykSzwer what about this? As I wrote  here and on Slack, I see no difference when the line `Session.Current.CalculatePatchAndPushOnWebSocket();` is removed.

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -142,7 +144,14 @@ void Handle(Input.CompositionsFolderApplyNow action)
 
         void Handle(Input.RecreateFilesFromDb action)
         {
-            if(!string.IsNullOrEmpty(FileSystemHelper.RegisteredCompositionsFolder))
+            if (!File.Exists(FileSystemHelper.RegisteredCompositionsFolder))
```
warpech: Shouldn't this use `RegisteredCompositionsFolder`?

src/BlendingEditor/Website/ViewModels/SettingsPage.json.cs
```diff
@@ -142,7 +144,14 @@ void Handle(Input.CompositionsFolderApplyNow action)
...
+            {
+                // Invoke this method to recalculate CompositionsFolderStatus value and send patch to the client
+                Session.Current.CalculatePatchAndPushOnWebSocket();
+                return;
+            }
+
+            if (!string.IsNullOrEmpty(FileSystemHelper.RegisteredCompositionsFolder))
```
warpech: Shouldn't this use `RegisteredCompositionsFolder`?

Issue #365, opened 2018-11-28
It's possible to ruin entire settings page with incorrect path


PatrykSzwer:
Right now you're able to input as compositions folder path that has illegal characters which will basically block entire settings page from loading:
Example string that I put there: `.(>.!=)`

Exception:
```
PalindromConnectionError: Server error
	System.Exception: ScErrDataBindingForJsonException (SCERR14012): An exception occurred when a databinding for a property in json was used. Property 'CompositionsFolderStatus' in 'MasterWrapper.LayoutPage.Json.ViewKeeper.KeptViews.KeptViewsElement.MasterPage.SettingsPage'.
	Version: 2.4.990.191.
	Help page: https://docs.starcounter.io/?q=SCERR14012. ---> System.ArgumentException: Illegal characters in path.
	   at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
	   at System.IO.Path.Combine(String path1, String path2)
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.GetAbsoluteFolderPath(String relativePath)
	   at BlendingEditor.Website.ViewModels.SettingsPage.get_CompositionsFolderFullPath()
	   at BlendingEditor.Website.ViewModels.SettingsPage.get_CompositionsFolderStatus()
	   at lambda_method(Closure , Json )
	   --- End of inner exception stack trace ---
	   at lambda_method(Closure , Json )
	   at Starcounter.Templates.Property`1.CheckAndSetBoundValue(Json parent, Boolean addToChangeLog) in c:\Git\Starcounter.XSON\src\Starcounter.XSON\Templates\Foundation\Property.cs:line 238
	   at Starcounter.Json.CheckOneTemplate(Json parent, TValue template, ChangeLog changeLog) in c:\Git\Starcounter.XSON\src\Starcounter.XSON\Class Json\Json.Object.ChangeLog.cs:line 444
	   at Starcounter.Json.CollectChanges(ChangeLog changeLog) in c:\Git\Starcounter.XSON\src\Starcounter.XSON\Class Json\Json.Object.ChangeLog.cs:line 323
	   at Starcounter.Json.RetriableCollectChanges(ChangeLog changeLog, Boolean retryOnError) in c:\Git\Starcounter.XSON\src\Starcounter.XSON\Class Json\Json.Object.ChangeLog.cs:line 300
[...]

PR #366, opened 2018-11-28
Fix for illegal characters in path

PatrykSzwer:
For https://github.com/Starcounter/Blending/issues/365


src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -50,14 +50,14 @@ public static string GetAbsoluteFolderPath(string relativePath)
...
-            string folderPath = Path.Combine(rootPath, relativePath);
             try
             {
+                string folderPath = Path.Combine(rootPath, relativePath);
                 return Path.GetFullPath(folderPath);
             }
             catch (Exception)
             {
-                return folderPath;
+                return rootPath + "\\" + relativePath;
```
warpech: I'd rather use `System.IO.Path.DirectorySeparatorChar` for future compatibility with Linux.

Can you pls add a comment above the method that says this method is will return a pseudo path even if it does not exist or contains incorrect characters?

PatrykSzwer: Done.

PR #367, opened 2018-11-28
Remove visual editor

### LINKED ISSUE #352: Remove the right "Visual editor" from default UI
Remove the right side "Visual editor". It gets in the way really and isn't helpful at least to me. I don't think this is something that we can change someone's mind about. It's one of the things where "I just don't like it" is a good argument. Maybe we need to vote. Also, now when you uncheck it out, the ace-editor doesn't know the window was resized, so it keeps wrapping the code early (before reaching full width). You need to resize the whole window to trigger recalcuation. 

Moved from https://github.com/Starcounter/starcounter-layout-html-editor/issues/44

alshakero:
fix #352


warpech:
I thought we agreed on **hiding** the visual editor by default, not removing it completely. I missed that #352 says "remove".

However, I am cool with removing it. Then, not long from now, we will have a simple editor for dummies and the advanced code editor.

@tomalec are you cool with it, too?

Issue #368, opened 2018-11-28
Importing composition from .json file without html value and file representation results in error


PatrykSzwer:
When investigating this one https://github.com/Starcounter/Blending/issues/334 I found out that when there is a composition in .json file without Html and there is no file for that composition it results in further problems, because there will be composition in database that `.Html` value is `null`.

Related to: https://github.com/Starcounter/Blending/issues/309

To invoke that problem you just need to click `Recreate HTML files from database` when there is incorrect composition in DB.

Exception that you will see:
```
System.NullReferenceException: Object reference not set to an instance of an object.
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.NormalizeLineEndings(String text, String lineEnding)
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.WriteToFileIfDifferent(String path, String content, Boolean force)
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.CopyFromCacheToFile(HtmlViewComposition composition, Boolean force)
	   at BlendingEditor.Composition.Helpers.FileSystemHelper.CopyFromCacheToFiles()
	   at BlendingEditor.Website.ViewModels.SettingsPage.Handle(RecreateFilesFromDb action)
	   at lambda_method(Closure , Json , Input`1 )
	   at Starcounter.XSON.Internal.XSONInternals.RunWithinApplication(String appName, Action action) in c:\Git\Starcounter.XSON\src\Starcounter.XSON\Internal\XSONInternals.cs:line 41
	   at Starcounter.Templates.Property`1.ProcessInput(Json parent, T newValue) in c:\Git\Starcounter.XSON\src\Starcounter.XSON\Templates\Foundation\Property.cs:line 340
	   at Starcounter.XSON.JsonPatch.DefaultPatchHandler.ParseAndProcess(TLong property, Json parent, String valueStr) in c:\Git\Starcounter.XSON\src\Starcounter.XSON\Sessions\JSON-Patch\DefaultPatchHandler.cs:line 147
	   at Starcounter.Internal.TransactionManager.Scope(TransactionHandle handle, Action action) in C:\tc_work\sc-19683\Level1\src\Starcounter\TransactionManager.cs:line 557
	   at Starcounter.Transaction.Scope(Action action) in C:\tc_work\sc-19683\Level1\src\Starcounter\Transaction.cs:line 233
[...]

PR #369, opened 2018-11-28
Fix unexpected exception with composition without html value in database

PatrykSzwer:
For #368


src/BlendingEditor/Composition/Helpers/FileSystemHelper.cs
```diff
@@ -176,7 +176,8 @@ public static void CopyFromCacheToFiles()
                 }
             }
 
-            foreach (HtmlViewComposition composition in compositions)
+            // Check in case of invalid (without Html) composition in the database. 
+            foreach (HtmlViewComposition composition in compositions.Where(x => x.Html != null))
```
warpech: I'd rather add it at line 159:

```c#
IEnumerable<HtmlViewComposition> compositions = HtmlViewCompositionHelper.GetAll().Where(x => x.Html != null);
```

It would be cleaner, especially given that there are two foreaches in this method.

PatrykSzwer: Done,  I also added try-catch block for recreating files so it will not be able to throw unexpected exception, instead it will be shown on the UI what happened.

warpech: I appreciate the message a lot! But I think it's usability is close to zero if it appears at the top of the page...

Could we reserve `Message`, `IsErrorMessage` for Import/Export result, and create a separate message properties for `Recreate HTML files from database`?

PatrykSzwer: I'll try to do something about that.

PatrykSzwer: Done.

src/BlendingEditor/wwwroot/BlendingEditor/views/SettingsPage.html
```diff
@@ -230,39 +239,48 @@ <h5>Are you sure you want reset to defaults?</h5>
...
                 <div class="uni-alert-danger">
-                    <slot name="blendingeditor/settings-folder-message-danger"></slot>
+                    <slot name="blendingeditor/settings-composition-alert-danger"></slot>
                 </div>
                 <div class="uni-alert-success">
-                    <slot name="blendingeditor/settings-folder-message-success"></slot>
-                </div>
-                <div class="uni-alert-info">
-                    <slot name="blendingeditor/settings-folder-message-info"></slot>
+                    <slot name="blendingeditor/settings-composition-alert-success"></slot>
```
warpech: Much better! I would like to make it even more granular, but I've added a separate issue to unblock this PR.

src/BlendingEditor/Website/ViewModels/SettingsPage.json
```diff
@@ -4,6 +4,8 @@
 
   "Message$": "",
   "IsErrorMessage$": false,
+  "CompositionsMessage$": "",
+  "IsCompositionsErrorMessage$": false,
```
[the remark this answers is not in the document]
PatrykSzwer: I basically copied messages from above but now I see that those 2 are actually being changed on client side trough script.

Issue #338, opened 2018-11-23
Store file names in the database, allow for custom or short file names, etc


warpech:
In level1, I recently added the `HtmlFilePath` property to `HtmlViewComposition` that should allow to have arbitrary, not generated file names for HTML custom compositions.

BlendingEditor should be changed to make use of that.

warpech:
I will take it with high prio because more and more people are reporting problems to checkout files with long names.

PatrykSzwer:
@warpech I'll assign this to you since you already said that you will take it. Unassign yourself if you disagree.

warpech:
Implemented in: https://github.com/Starcounter/Blending/pull/372

PR #372, opened 2018-11-28
add UI, storage, export and import for custom file names (#338)

warpech:


warpech:
@mmnosek, @Mihaiii this is probably the solution for the files that you can't check out.

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
```diff
@@ -36,7 +36,44 @@ public string FilePath
...
             {
-                var path = this.Data != null ? FileSystemHelper.GetAbsoluteHtmlFilePath(this.Data) : null;
+                var path = this.Data != null && FileSystemHelper.IsRegisteredCompositionsFolder ? HtmlViewCompositionHelper.GetHtmlFileName(this.Data) : null;
+                return path;
+            }
+            set
+            {
+                var newFileName = value;
+
+                if (newFileName == "") //this means "restore the initial file name"
```
PatrykSzwer: Use `String.Empty` instead of "".

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
```diff
@@ -36,7 +36,44 @@ public string FilePath
         {
```
PatrykSzwer: I know its not part of changes in this PR but in this file at the top there is field:
`        private string OriginalHtml;` which should be lower-case. Could you change that?

src/BlendingEditor/Website/ViewModels/CompositionPage.json.cs
```diff
@@ -36,7 +36,44 @@ public string FilePath
...
+                try
+                {
+                    string newPath = FileSystemHelper.FilePathToAbsoluteFilePath(newFileName);
+                    FileSystemHelper.MoveFileOrDeleteIfExists(this.Data, newPath);
+                    Db.Transact(() =>
+                    {
+                        this.Data.HtmlFilePath = value;
+                    });
+                }
+                catch (Exception)
```
PatrykSzwer: Are you sure you want to just silently catch exception here without doing anything? I would suggest showing here some error message.

Issue #247, opened 2018-06-14
Delete relatives after deleting Layout


kegor:
### Steps to reproduce
1. Delete Layout.
2. Check Areas in the database.

### Expected result
They have Deleted field equals to true.

### Actual result
They have Deleted field == false.

##### Starcounter version 
2.4.990.103

##### Involved apps' version
 - Starcounter/BlendingEditor@6.3.0
 - Starcounter/BlendingProvider@6.3.0

##### Browsers affected

- [x] Chrome
- [x] Firefox

### Notes
Originally reported in https://github.com/Starcounter/Blending/pull/233#pullrequestreview-128799640.

kegor:
This might be implemented using https://docs.starcounter.io/v/2.4/topic-guides/database/referential-integrity-and-constraints#ondelete.

PatrykSzwer:
For some reason github automatically closed this issue, when It's still not done.

PatrykSzwer:
>This might be implemented using https://docs.starcounter.io/v/2.4/topic-guides/database/referential-integrity-and-constraints#ondelete.

I think it cant be done this way, because deleting function for layout is not calling `.Delete()`, its just setting its `Deleted` property to true.

PatrykSzwer:
Closing as done in #376

PR #376, opened 2018-11-30
Deleting layout relatives

PatrykSzwer:
For https://github.com/Starcounter/Blending/issues/247

So guys, basically I'm just marking every layout relatives as deleted, but those are not deleted from database as entities (ever?). I'm not sure if should I maybe just delete them instead of marking as Deleted? @un-tone @kegor maybe you will know better which one I should delete and how and which should I mark as deleted and why.


warpech:
I know it is weird with this `Deleted` property. For now, I would say:

- if there is a `Deleted` property in the db class, use it
- otherwise `Delete()`

warpech:
**Before merging please check if the tests are passing**

The first build failed: https://teamcity.starcounter.org/viewLog.html?buildId=81255&buildTypeId=Apps_Blending_BlendingMasterWithStarcounterDevelop24

I am running a second build: https://teamcity.starcounter.org/viewLog.html?buildId=81296&

PatrykSzwer:
I run tests again and they passed (flaky tests). Merging.

PR #377, opened 2018-12-03
Use the unauthorized page that is provided by Authorization library

warpech:
Use the unauthorized page that is probided by Authorization library, instead of a custom DenyPage.

Allows to implement: https://github.com/Starcounter/HeadsOmni/issues/293


warpech:
Yes, it requires https://github.com/Starcounter/SignIn/pull/180.

Thx for the review!

PR #378, opened 2018-12-03
Make Composition Editor searchable by view-urls too

### LINKED ISSUE #286: Composition editor starcounter-include dropdown should allow to filter by View URI
Currently, the dropdown only allows to filter by CSS selector of `starcounter-include`, which is not very useful. Filtering by typing an app name or HTML file name would be more useful.

alshakero:
I noticed that without highlighting it takes much longer to find the correct partial with a short query. When I type `topbar` I see a list of 5 elemetns that I have to parse yet again. With highlighting it's better UX. 

How it looks now:
compoeditorsearch

Fix #286.


src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -124,8 +123,8 @@
                              color: grey;
                          }
                      </style>
-                     <p class="indented">[[item.indent]][[item.name]]</p>
-                     <p class="indented grey">[[item.indent]]View: [[item.view]]</p>
+                     <p class="indented" inner-h-t-m-l="[[item.indent]][[item.highlightedName]]"></p>
```
tomalec: Why this trick is needed?

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -385,12 +390,41 @@
...
+                    /* filter elements and highlight matching substrings in names and views */
+                    event.target.filteredItems = clone.filter(item => {
+                        /* we need a regex because:
+                            1. greediness
+                            2. search is case-insensetive
+                            3. to use the match as the replacement value, hence preserving the case of the match,
+                               even though the search is case-insensetive and `query` is in lowercase.
+                        */
+                        if(item.name.toLowerCase().includes(query)) {                            
+                            item.highlightedName = item.highlightedName.replace(new RegExp(query, "ig"), match => `<span style="background: lightblue; border-radius: 3px">${match}</span>`);
```
tomalec: I'd rather use a class, not to hardcode style rules here

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -385,12 +390,41 @@
...
                             const depthSpaces = "   ".repeat(calculateDepth(elem));
                             const viewHtml = elem.viewModel[elem.compositionProvider].PartialId;
-
-                            knownSelectors.push({ name: selector, indent: depthSpaces, view: viewHtml });
+                            knownSelectors.push({ name: selector, highlightedName: selector, highlightedView: viewHtml, indent: depthSpaces, view: viewHtml });
                         }
                     });
                     return knownSelectors;
                 },
+                filter(event) {
```
tomalec: Is highlighting the entered query, the only reason not to use built-in search?

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: Because I'm inserting HTML not just plain test.

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: I tried. But this goes very deep in the shadow of `vaadin-list-element` then `iron-list-item` etc.

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: This and highlighting.

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
tomalec: Can't you add those rules in template? as in https://vaadin.com/components/vaadin-combo-box/html-examples/combo-box-styling-demos#using-focused-and-selected-flags

Can you then at least use some Vaadin custom properties? so the `lightblue` would addapt if you change Vaadin props.

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
alshakero: I misunderstood your questions. The reasons are:

1. Searching for both view URL and selector.
2. Highlighting.

tomalec: Ok, I get the value of highlighting, but do we really want to be able to search by selector?, especially that no, it does not except "any selector matching given sc-include", but it's searching among selectors provided by us.

alshakero: I'm cool with anything. Would you like me to remove showing selectors and using the default search?

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
warpech: Seems legit to me, an alternative would be a `dom-repeat` for every character in the string :D

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
warpech: Yeah, search for selectors seems quite pointless. Search for main and attached view URIs makes the only sense.

warpech:
On top to what was said, I think that view URIs should be presented with big font and selectors not at all, or with a small font.

tomalec:
Ok, so do we all agree that:
1. we will remove searching dropdown results by selector,
2. Remove selector at all,
3. Keep non-vaadin filtering just to have highlighting,
4. Keep highlighting rules set as a class in style element of a dropdown item template?

Issue #22, opened 2018-01-08
Add an icon to each blending point that opens the editor for that blending point


PatrykSzwer:
_Issue moved from CompositionEditor repository: https://github.com/Starcounter/CompositionEditor/issues/7_

Add an icon to each blending point that opens the editor for that blending point

The icon can be added as CSS `::before`

warpech:
I will do some research for it.

warpech:
This depends on https://github.com/Starcounter/Blending/issues/23 being done first.

tomalec:
I consider adding a simple way to pick a blending point to edit its composition a requirement of https://github.com/Starcounter/Blending/issues/149

I thought about adding a button "Pick a view",  when cliecked or when simple editor is opened, let's you just click anywhere on the page to select the leaf `starcounter-include` which will be highlighted. Then you can in folder-like manner go up to its parent views.

warpech:
It was discussed today by @tomalec and me that there will be a Launchpad icon that toggles composition editing ON and OFF. When composition editing is ON, there is a floating toolbar presented on the screen.

The floating toolbar contains:
1. a dropdown to pick blending points. All blending points are listed as a tree list inside the dropdown
  - when you highlight an item in the dropdown, the relevant blending point should be highlighted in the document
  - possibly in future there can also be a button to move up to the from the current blending point to its parrent. But we skip that for now, because it duplicates dropdown in a way
2. an icon to select a blending point by point & click. This method selects the deepest (leaf) blending point under the mouse cursor
3. a simple editor for the selected blending point (main issue here: https://github.com/Starcounter/Blending/issues/149)
4. a button to open the advanced editor in a popup

warpech:
It was agreed on Slack that @alshakero will take the point 0 of https://github.com/Starcounter/Blending/issues/22#issuecomment-442866992. I can help to make the icon that turns custom composition editing ON and OFF.

I proposee that @alshakero also takes points 1 and 4.

@tomalec takes points 2 and 3.

alshakero:
It was agreed in a meeting between @tomalec and I that we use a sticky top bar instead of a floating snapping draggable. The UX sounds better with the sticky bar and there is more area in the top bar than a floating toolbox. 

The PR is here https://github.com/Starcounter/Blending/pull/391, it's still WIP.

PR #382, opened 2018-12-04
 Add `blending-point-picker`, add travis setup for multiple elements

tomalec:
- Add first version of `blending-point-picker` #22
- setup `wctall` script to be able to test many elements in TravisCI


src/BlendingEditor/wwwroot/BlendingEditor/elements/blending-point-picker/test/spec.html
```diff
@@ -0,0 +1,73 @@
...
+</body>
+<script>
+    var juicyElement;
+    describe('blending-point-picker', function() {
+
+        beforeEach(function() {
+            [picker, blendingPoint, anotherBlendingPoint] = fixture('element');
+        });
+
+        it('should be upgraded element', function() {
```
alshakero: This test will never fail. `<undefined-element> is never instanceof HTMLUnknownElement` and is always `HTMLElement`.

wctall.js
```diff
@@ -0,0 +1,25 @@
+#! /usr/bin/env node
+var shell = require("shelljs");
+
+const [,, ... args] = process.argv;
+const package = require("./package.json");
+
+let failure = false;
+const results = package.config.webcomponents.map((path)=>{
+    const result = shell.exec(`cd ${path} && bower instal && wct ${args.join(' ')}`).code;
```
alshakero: Typo
```suggestion
    const result = shell.exec(`cd ${path} && bower install && wct ${args.join(' ')}`).code;
```

wctall.js
```diff
@@ -0,0 +1,25 @@
+#! /usr/bin/env node
+var shell = require("shelljs");
+
+const [,, ... args] = process.argv;
```
alshakero: What's the reason for adding two empty elements in the array top?

alshakero:
Addressed the review comments and merge to base my changes of the testing apparatus you created.

Issue #383, opened 2018-12-04
Adopt to changes required by Orchestrator


warpech:
@PatrykSzwer pls adopt to changes required by @alemoi in https://github.com/Starcounter/Starcounter.Orchestrator/issues/2#issuecomment-440356082

alemoi:
@warpech I asked @PatrykSzwer about many other apps that are using `Blender` API directly, instead of `blend.json` and he is not sure. So I need to ask you about them :)

Here is the list of apps that I found using `Blender` API:
```
Starcounter/Website 
Starcounter/Timeline
Starcounter/Barcodes
Starcounter/inside.starcounter.com
Starcounter/Notes 
Starcounter/SimpleTimeline
Starcounter/Cashier 
Starcounter/BlendingTutorial
Starcounter/SignIn
Starcounter/Skyper 
Starcounter/SyncEngine
Starcounter/Blending
Starcounter/Timereport
Starcounter/Starcounter.Authorization
```

PR #385, opened 2018-12-05
Adapt changes required by Orchestrator

PatrykSzwer:
For https://github.com/Starcounter/Blending/issues/383


alemoi:
> This should be merged after branches https://github.com/Starcounter/level1/tree/mapping-separate-blender and https://github.com/Starcounter/level1/tree/develop-separate-blender are merged in Level1.

I just have merged those in Level1, so merging this branch as well then. CC @warpech @PatrykSzwer

PR #391, opened 2018-12-05
WIP: Add draggable quick picker

alshakero:
Implements point 0 of https://github.com/Starcounter/Blending/issues/22#issuecomment-442866992

How to try it:

1. Run this app and SignIn
2. SignIn.
3. Go to http://localhost:8080/BlendingEditor/layoutgroups (arbitrary). 
4. Drag it around.


warpech:
I played with it, looks close to what I expected! Few comments:

1. please use `move` mouse instead of `grab`. Explanation: https://ux.stackexchange.com/questions/75816/when-to-use-grab-vs-move-cursor
2. the whole padding area of the toolbar should be draggable
3. the icon  to me represents "photo adjustment" tools. I think it has no place here. If you do point 3, a separate icon is not needed to represent draggability.
4. dragging the toolbar should not select the text on the web page
5. right now, as I drag, the toolbar changes the appearance to transparent window. I think that is unneeded.
6. when I drag the toolbar to the right hand side edge of the window, a part of the toolbar is hidden behind my window scrollbar
7. the icon to open the advanced editor is not very intuitive. Perhaps this would be better?  (taken from Chrome DevTools)

src/BlendingEditor/wwwroot/BlendingEditor/elements/compositioneditor-inserter/compositioneditor-inserter.html
```diff
@@ -2,7 +2,7 @@
 <link rel="import" href="/sys/juicy-highlight/juicy-highlight.html" />
 
 <dom-module id="compositioneditor-inserter">
-    <template>
+    <template>        
         <compositioneditor-hot-key keys="Ctrl+E" on-hotkeydown="_showPanel"></compositioneditor-hot-key>
```
warpech: Please remove `compositioneditor-hot-key`. This PR removes the previous method of opening the popup composition editor.

src/BlendingEditor/wwwroot/BlendingEditor/elements/starcounter-layout-html-editor/starcounter-layout-html-editor.html
```diff
@@ -246,23 +202,29 @@
...
+                    if(!selector) {
+                        return;
+                    }
+                    const partial = documentContext.querySelector(selector);
+                    partial && this._blendingPointPicked(partial);
                 },
                 _blendingPointClicked(event){
-                    this.partial = event.detail;
+                    const partial = event.detail;
+                    const selector = `starcounter-include[slot="${partial.getAttribute('slot')}"]`;
```
warpech: What about `starcounter-include`s that do not have a `slot` attribute?

